PR #6, opened 2025-08-05
chore: Fixes before release

bracesproul:
Changes:
- Adds CI
- Improves formatter/linter scripts
- Adds separate build for examples so its not included in released code
- changed version to `0.0.0` in preparation for launch
- added MIT license

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@ Using an LLM to call tools in a loop is the simplest form of an agent. This arch
 
 `deepagents` is a TypeScript package that implements these in a general purpose way so that you can easily create a Deep Agent for your application.
 
+> ![TIP]
+> Looking for the Python version of this package? See [here: hwchase17/deepagents](https://github.com/hwchase17/deepagents)
+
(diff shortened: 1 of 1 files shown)


src/subAgent.ts
```diff
@@ -34,12 +33,21 @@ const BUILTIN_TOOLS: Record<string, StructuredTool> = {
...
  * Uses Command for state updates and navigation between agents.
  */
-export function createTaskTool(
-  subagents: SubAgent[],
-  tools: Record<string, StructuredTool> = {},
-  model = getDefaultModel(),
-  stateSchema = DeepAgentState,
-) {
+export function createTaskTool<
+  StateSchema extends z.ZodObject<any, any, any, any, any>,
```
Palashio, line 37: i'm seeing a lot of changes with this quadruple any structure? curious what the reasoning here is?

bracesproul: `z.ZodObject` requires genetics to be passed in describing the contents of the zod object. we dont care what the contents are (hence the `any`'s), as long as they're inside a zod object

PR #19, opened 2025-08-25
feat: add config

Palashio:
adds interrupt config

### DIFF
--- a/src/graph.ts
+++ b/src/graph.ts
@@ -16,6 +16,7 @@ import type { CreateDeepAgentParams } from "./types.js";
 import type { StructuredTool } from "@langchain/core/tools";
 import { z } from "zod";
 import { DeepAgentState } from "./state.js";
+import { createInterruptHook } from "./interrupt.js";
 
 /**
  * Base prompt that provides instructions about available tools
@@ -59,6 +60,7 @@ export function createDeepAgent<
     model = getDefaultModel(),
     subagents = [],
     postModelHook,
+    interruptConfig = {},
   } = params;
 
   const stateSchema = params.stateSchema
@@ -91,12 +93,29 @@ export function createDeepAgent<
     ? instructions + BASE_PROMPT
     : BASE_PROMPT;
 
+  // Should never be the case that both are specified
+  if (postModelHook && Object.keys(interruptConfig).length > 0) {
+    throw new Error(
+      "Cannot specify both postModelHook and interruptConfig together. " +
+        "Use either interruptConfig for tool interrupts or postModelHook for custom post-processing.",
+    );
+  }
+
+  let selectedPostModelHook: typeof postModelHook;
+  if (postModelHook !== undefined) {
+    selectedPostModelHook = postModelHook;
+  } else if (Object.keys(interruptConfig).length > 0) {
+    selectedPostModelHook = createInterruptHook(interruptConfig);
+  } else {
+    selectedPostModelHook = undefined;
+  }
+
   // Return createReactAgent with proper configuration
   return createReactAgent({
     llm: model,
     tools: allTools,
     stateSchema,
     messageModifier: finalInstructions,
-    postModelHook,
+    postModelHook: selectedPostModelHook,
   });
 }
--- a/src/graph.ts
+++ b/src/graph.ts
@@ -12,7 +12,7 @@ import { createReactAgent } from "@langchain/langgraph/prebuilt";
 import { createTaskTool } from "./subAgent.js";
 import { getDefaultModel } from "./model.js";
 import { writeTodos, readFile, writeFile, editFile, ls } from "./tools.js";
-import type { CreateDeepAgentParams } from "./types.js";
+import type { CreateDeepAgentParams, PostModelHook } from "./types.js";
 import type { StructuredTool } from "@langchain/core/tools";
 import { z } from "zod";
 import { DeepAgentState } from "./state.js";
@@ -119,3 +119,4 @@ export function createDeepAgent<
     postModelHook: selectedPostModelHook,
   });
 }
+
--- a/src/graph.ts
+++ b/src/graph.ts
@@ -119,5 +119,3 @@ export function createDeepAgent<
     postModelHook: selectedPostModelHook,
   });
 }
-
-
--- a/src/graph.ts
+++ b/src/graph.ts
@@ -101,7 +101,7 @@ export function createDeepAgent<
     );
   }
 
-  let selectedPostModelHook: typeof postModelHook;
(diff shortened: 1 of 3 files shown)


src/graph.ts
```diff
@@ -91,12 +93,29 @@ export function createDeepAgent<
...
 
+  // Should never be the case that both are specified
+  if (postModelHook && Object.keys(interruptConfig).length > 0) {
+    throw new Error(
+      "Cannot specify both postModelHook and interruptConfig together. " +
+        "Use either interruptConfig for tool interrupts or postModelHook for custom post-processing.",
+    );
+  }
+
+  let selectedPostModelHook: typeof postModelHook;
```
bracesproul: could we create a `PostModelHook` type in `types.ts`, use it in the `CreateDeepAgentParams` interface & import to use here.

```typescript
export type PostModelHook = (
    state: DeepAgentStateType,
    model: LanguageModelLike,
  ) => Promise<DeepAgentStateType>;
```

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
...
+export interface HumanInterrupt {
+  action_request: ActionRequest;
+  config: HumanInterruptConfig;
+  description: string;
+}
+
+export interface HumanResponse {
+  type: "accept" | "edit" | "ignore";
+  args?: ActionRequest;
+}
```
bracesproul: these can be imported from `@langchain/langgraph/prebuilt` i believe

src/types.ts
```diff
@@ -43,9 +44,13 @@ export type LanguageModelLike = Runnable<
   LanguageModelOutput
 >;
 
-/**
- * Parameters for createDeepAgent function with TypeScript types
- */
+export interface HumanInterruptConfig {
```
bracesproul: can also be imported from prebuilt

bracesproul: and i think it's snake case, but import should confirm that

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export interface HumanInterrupt {
+  action_request: ActionRequest;
+  config: HumanInterruptConfig;
+  description: string;
+}
+
+export interface HumanResponse {
+  type: "accept" | "edit" | "ignore";
```
bracesproul: no `"response"`?

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
...
+   *   messagePrefix: Optional message prefix for interrupt descriptions
+   */
+
+  return async function interruptHook(
+    state: DeepAgentStateType,
+  ): Promise<DeepAgentStateType> {
+    const messages = state.messages || [];
+    if (!messages.length) {
+      return state;
+    }
```
bracesproul: will this act like a node? if so, you shouldn't need to return the state since it's checkpointed

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    const lastMessage = messages[messages.length - 1] as AIMessage;
+    if (!lastMessage.tool_calls || !lastMessage.tool_calls.length) {
+      return state;
+    }
```
bracesproul: don't blindly assume it's an ai message
```suggestion
    const lastMessage = messages[messages.length - 1];
    if (!isAIMessage(lastMessage) || !lastMessage.tool_calls || !lastMessage.tool_calls.length) {
      return state;
    }
```

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    // Separate tool calls that need interrupts from those that don't
+    const interruptToolCalls: any[] = [];
+    const autoApprovedToolCalls: any[] = [];
```
bracesproul: ahhhhhh `any`'s jump scare

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
...
+          action: toolName,
+          args: toolArgs,
+        },
+        config: toolConfig,
+        description: description,
+      };
+      requests.push(request);
+    }
+
+    const responses: HumanResponse[] = await interrupt(requests);
```
bracesproul: nit, interrupt takes generics
```suggestion
    const responses = await interrupt<HumanInterrupt[], HumanResponse[]>(requests);
```

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
...
+          name: toolCall.name,
+          args: edited.args,
+          id: toolCall.id,
+        };
+        approvedToolCalls.push(newToolCall);
+      } else if (response.type === "ignore") {
+        continue;
+      } else {
+        throw new Error(`Unknown response type: ${response.type}`);
+      }
```
bracesproul, line 100: support response type?

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    const updatedLastMessage = {
+      ...lastMessage,
+      tool_calls: approvedToolCalls,
+    };
+
+    const updatedMessages = [...messages.slice(0, -1), updatedLastMessage];
+
+    return { ...state, messages: updatedMessages as typeof messages };
```
bracesproul: same as above, shouldn't need to include state if it's acting as a node
```suggestion
    return { messages: updatedMessages as typeof messages };
```

bracesproul: and if so, you should only need to return the new `updatedLastMessage`, not the entire messages list (since it has the same id as before the update, it should take precedence in the reducer)

src/interrupt.ts
```diff
@@ -0,0 +1,119 @@
+import type { HumanInterruptConfig } from "./types.js";
+import type { DeepAgentStateType } from "./types.js";
+import { interrupt } from "@langchain/langgraph";
+import { AIMessage } from "@langchain/core/messages";
+
+export type ToolInterruptConfig = Record<string, HumanInterruptConfig>;
```
bracesproul: move this to `types.ts`?

bracesproul:
@open-swe I left a detailed review above with change requests. Could you please resolve all of the comments?

src/interrupt.ts
```diff
@@ -0,0 +1,113 @@
...
+        };
+        approvedToolCalls.push(newToolCall);
+      } else if (response.type === "ignore") {
+        continue;
+      } else if (response.type === "response") {
+        // continue here since it's a direct feedback
+        continue;
+      } else {
+        throw new Error(`Unknown response type: ${response.type}`);
+      }
```
bracesproul: discussed in slack, lets throw err here and say it's not supported
```suggestion
      } else {
        throw new Error(`Unknown response type: ${response.type}`);
      }
```

bracesproul (approved):

PR #26, opened 2025-08-27
feat: Expose context schema arg

bracesproul:


src/types.ts
```diff
@@ -53,12 +57,19 @@ export type ToolInterruptConfig = Record<string, HumanInterruptConfig>;
...
+    | AnyAnnotationRoot
+    | InteropZodObject = AnyAnnotationRoot,
 > {
   tools?: StructuredTool[];
   instructions?: string;
   model?: LanguageModelLike;
   subagents?: SubAgent[];
   stateSchema?: StateSchema;
+  /**
+   * An optional schema for the context.
```
Palashio: ✏️ unnecessary comment?

src/types.ts
```diff
@@ -15,8 +15,12 @@ import type { StructuredTool } from "@langchain/core/tools";
 import type { DeepAgentState } from "./state.js";
 import { z } from "zod";
 import { Runnable } from "@langchain/core/runnables";
+import { AnnotationRoot } from "@langchain/langgraph";
+import { InteropZodObject } from "@langchain/core/utils/types";
 import type { HumanInterruptConfig } from "@langchain/langgraph/prebuilt";
 
+export type AnyAnnotationRoot = AnnotationRoot<any>;
```
Palashio, line 22: ✏️ any

bracesproul: any in this case is okay. we need to pass through a generic, and there isn't a good way (or a need) to know at runtime what the shape will be. this is just used to constrain the generic we expose

PR #28, opened 2025-08-28
feat: improve interrupt

Palashio:
- adds the improvements to interrupt but in javascript, same functionality for python found here: https://github.com/langchain-ai/deepagents/pull/79

### DIFF
--- a/src/interrupt.ts
+++ b/src/interrupt.ts
@@ -96,8 +96,8 @@ export function createInterruptHook(
       description: description,
     };
 
-    const responses: HumanResponse[] = await interrupt([request]);
-
+    const res: HumanResponse | HumanResponse[] = await interrupt([request]);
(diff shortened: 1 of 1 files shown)


src/interrupt.ts
```diff
@@ -56,48 +68,63 @@ export function createInterruptHook(
...
+        action: toolName,
+        args: toolArgs,
+      },
+      config: typeof toolConfig === "object" ? toolConfig : defaultToolConfig,
+      description: description,
+    };
+
+    const responses: HumanResponse[] = await interrupt([request]);
+
+    if (responses.length !== 1) {
```
bracesproul: nit, always ensure the response is an array
```suggestion
    const res: HumanResponse | HumanResponse[] = await interrupt([request]);
    const responses = Array.isArray(res) ? res : [res];

    if (responses.length !== 1) {
```

PR #36, opened 2025-09-26
fix: migrate to pnpm, secure workflows

christian-bromann:
This patch contains some general maintenance work:

- migrated to pnpm from outdated Yarn v1
- adding a dependabot for workflow and package version updates
- use shasums for workflows as it is more secure


christian-bromann:
replaced by https://github.com/langchain-ai/deepagentsjs/pull/87

PR #37, opened 2025-09-29
feat: Add a name property value to ToolMessage

wuchangming:
Add an explicit name attribute to tools and guarantee a non-undefined fallback when absent; this is fully backward-compatible.


christian-bromann (approved):
LGTM 👍

PR #43, opened 2025-11-04
feat: add pluggable filesystem

nhuang-lc:
- Move to using the pluggable filesystem

### DIFF
--- a/examples/research/research-agent.ts
+++ b/examples/research/research-agent.ts
@@ -65,7 +65,7 @@ const internetSearch = tool(
 
 const subResearchPrompt = `You are a dedicated researcher. Your job is to conduct research based on the users questions.
 
-Conduct thorough research and then reply to the user with a detailed answer to their question
+CConduct thorough research and then reply to the user with a detailed answer to their question
 
 only your FINAL answer will be passed on to the user. They will have NO knowledge of anything except your final message, so your final report should be your final message!`;
 
--- a/examples/backends/composite-backend.ts
+++ b/examples/backends/composite-backend.ts
@@ -97,9 +97,6 @@ async function main() {
     },
   );
 
-  console.log("Research Complete!");
-  console.log(`Thread ID: ${threadId}`);
-
   const threadId2 = uuidv4();
   await agent.invoke(
     {
--- a/examples/backends/filesystem-backend.ts
+++ b/examples/backends/filesystem-backend.ts
@@ -38,9 +38,7 @@ export const agent = createDeepAgent({
 });
 
 async function main() {
-  console.log(`Working directory: ${workspaceDir}\n`);
-
-  const result = await agent.invoke(
+  await agent.invoke(
     {
       messages: [
         new HumanMessage(
@@ -50,9 +48,6 @@ async function main() {
     },
     { recursionLimit: 50 },
   );
-
-  console.log("\n✅ Coding task complete!");
-  console.log(`\nFiles written to: ${workspaceDir}`);
 }
 
 if (import.meta.url === `file://${process.argv[1]}`) {
--- a/examples/backends/state-backend.ts
+++ b/examples/backends/state-backend.ts
@@ -54,21 +54,13 @@ export const agent = createDeepAgent({
 });
 
 async function main() {
-  const result = await agent.invoke(
+  await agent.invoke(
     {
       messages: [
         new HumanMessage("Research the latest trends in AI agents for 2025"),
       ],
     },
     { recursionLimit: 50 },
-  );
-
-  console.log("Research Complete!");
-  console.log("\nFiles created:");
-  console.log(
-    Object.keys(result.files || {})
-      .map((key) => `  - ${key}`)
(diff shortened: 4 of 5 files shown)


src/backends/filesystem.ts
```diff
@@ -0,0 +1,619 @@
...
+      if (relative.startsWith("..") || path.isAbsolute(relative)) {
+        throw new Error(`Path: ${full} outside root directory: ${this.cwd}`);
+      }
+      return full;
+    }
+
+    if (path.isAbsolute(key)) {
+      return key;
+    }
+    return path.resolve(this.cwd, key);
```
dqbd, line 86: I think even with `virtualMode: false` we do want to avoid resolving outside of `cwd`?

src/backends/filesystem.ts
```diff
@@ -0,0 +1,619 @@
...
+      try {
+        const fd = await fs.open(
+          resolvedPath,
+          fsSync.constants.O_RDONLY | (fsSync.constants.O_NOFOLLOW || 0),
+        );
+        content = await fd.readFile({ encoding: "utf-8" });
+        await fd.close();
+      } catch {
+        // Fallback to normal open if O_NOFOLLOW fails
+        content = await fs.readFile(resolvedPath, "utf-8");
```
dqbd, line 219: _question_: what's the point of `O_NOFOLLOW` if we're always retrying without the flag?

src/backends/protocol.ts
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * Different contexts build this differently:
+ * - Tools: Extract state via getCurrentTaskInput(config)
+ * - Middleware: Use request.state directly
+ */
+export interface StateAndStore {
+  /** Current agent state with files, messages, etc. */
+  state: any;
+  /** Optional BaseStore for persistent cross-conversation storage */
+  store?: any;
```
dqbd: _nit_: type as `BaseStore`?

src/backends/protocol.ts
```diff
@@ -0,0 +1,214 @@
...
+ * - state: Current agent state (with files, messages, etc.)
+ * - store: Optional persistent store for cross-conversation data
+ *
+ * Different contexts build this differently:
+ * - Tools: Extract state via getCurrentTaskInput(config)
+ * - Middleware: Use request.state directly
+ */
+export interface StateAndStore {
+  /** Current agent state with files, messages, etc. */
+  state: any;
```
dqbd: _nit_: expose as generic and/or type as `unknown`?

src/agent.ts
```diff
@@ -51,7 +57,19 @@ export interface CreateDeepAgentParams<
...
+  /**
+   * Optional backend for filesystem operations.
+   * Can be either a backend instance or a factory function that creates one.
+   * If provided, this will be used instead of useLongtermMemory.
+   * The factory receives a config object with state and store.
+   */
+  backend?:
+    | BackendProtocol
+    | ((config: { state: any; store?: BaseStore }) => BackendProtocol);
+  /**
```
dqbd: _nit_: since it's 1.x now, fine to just remove `useLongtermMemory` instead?

src/backends/composite.ts
```diff
@@ -0,0 +1,256 @@
...
+   *          prefix removed (but keeps leading slash).
+   */
+  private getBackendAndKey(key: string): [BackendProtocol, string] {
+    // Check routes in order of length (longest first)
+    for (const [prefix, backend] of this.sortedRoutes) {
+      if (key.startsWith(prefix)) {
+        // Strip full prefix and ensure a leading slash remains
+        // e.g., "/memories/notes.txt" → "/notes.txt"; "/memories/" → "/"
+        const suffix = key.substring(prefix.length);
+        const strippedKey = suffix ? "/" + suffix : "/";
```
dqbd, line 54: _note_: pretty sure that windows won't work, maybe good to add windows to CI matrix

src/backends/filesystem.ts
nhuang-lc: This is a design decision from Python (we also allow access outside of the cwd). I think this is reasonable (claude code can do things outside of cwd), why do you think we should avoid?

src/backends/filesystem.ts
nhuang-lc: Fixed this logic, use lstat to check any symlinks and don't fallback without the flag anymore.

examples/research/research-agent.ts
```diff
@@ -65,7 +65,7 @@ const internetSearch = tool(
 
 const subResearchPrompt = `You are a dedicated researcher. Your job is to conduct research based on the users questions.
 
-Conduct thorough research and then reply to the user with a detailed answer to their question
+CConduct thorough research and then reply to the user with a detailed answer to their question
```
christian-bromann, line 68:
```diff
- CConduct thorough research and then reply to the user with a detailed answer to their question
+ Conduct thorough research and then reply to the user with a detailed answer to their question
```

PR #53, opened 2025-11-12
fix: thread config options to subagents

hntrl:
* adds config option to subagent invocations
* updates langchain and removes note about summarization middleware
* updates params to use more appropriate BaseLanguageModel class


nhuang-lc (approved):
Nice thank you!

Issue #65, opened 2025-11-23
InvalidUpdateError when running research example - "LastValue can only receive one value per step


crazyyanchao:
**Description:**
I encountered an `InvalidUpdateError` when trying to run the research example at:
https://github.com/crazyyanchao/deepagentsjs/tree/main/examples/research

**Error Details:**
```
InvalidUpdateError: Invalid update for channel "todos" with values [[{"content":"保存用户原始问题到 question.txt","status":"completed"},{"content":"使用 research-agent 深入研究 '上下文工程' 的定义、历史、关键概念、应用和最佳实践。用中文研究并总结关键点，确保信息全面。","status":"completed"},{"content":"基于研究结果，撰写 final_report.md 的初稿，用中文写成结构化的 Markdown 报告，包括标题、部分、引用等。","status":"in_progress"},{"content":"使用 critique-agent 审阅 final_report.md，提供改进建议。","status":"pending"},{"content":"根据批判反馈，编辑 final_report.md 直到满意。","status":"pending"}],[{"content":"保存用户原始问题到 question.txt","status":"completed"},{"content":"使用 research-agent 深入研究 '上下文工程' 的定义、历史、关键概念、应用和最佳实践。用中文研究并总结关键点，确保信息全面。","status":"completed"},{"content":"基于研究结果，撰写 final_report.md 的初稿，用中文写成结构化的 Markdown 报告，包括标题、部分、引用等。","status":"completed"},{"content":"使用 critique-agent 审阅 final_report.md，提供改进建议。","status":"in_progress"},{"content":"根据批判反馈，编辑 final_report.md 直到满意。","status":"pending"}]]: LastValue can only receive one value per step.

Troubleshooting URL: https://docs.langchain.com/oss/javascript/langgraph/INVALID_CONCURRENT_GRAPH_UPDATE/

    at StreamManager.value [as start] (http://localhost:5410/_next/static/chunks/node_modules_%40langchain_langgraph-sdk_ab843275._.js:2349:43)
    at async Object.submit (http://localhost:5410/_next/static/chunks/node_modules_%40langchain_langgraph-sdk_ab843275._.js:2870:9)
```

**Steps to Reproduce:**
1. Navigate to the research example directory
2. Run the example as instructed
3. The error occurs during execution

**Expected Behavior:**
The research example should run successfully without InvalidUpdateError.

**Actual Behavior:**
[...]

PR #74, opened 2025-12-05
fix: concurrent graph update issue (Issues #52 + 65)

brettshollenberger:
## Fixes Issues

* https://github.com/langchain-ai/deepagentsjs/issues/52
* https://github.com/langchain-ai/deepagentsjs/issues/65

## Why

This was a sneaky bug. It's gonna be a long walk so bear with me.

The files channel defines a reducer via `withLangGraph()`, so why are all these bugs complaining about the `LastValue` reducer being used? `LastValue` is only supposed to be a fallback when no reducer is defined.

```typescript
  const meta = schemaMetaRegistry.get(zodSchemaObject);
  if (meta?.reducer) {
    return new BinaryOperatorAggregate(meta.reducer);  // Uses your reducer
  } else {
    return new LastValue();  // Fallback - only one value per step allowed
  }
```

But we DO have a reducer!

```typescript
  const FilesystemStateSchema = z3.object({
    files: withLangGraph(z3.record(...), {
      reducer: { fn: fileDataReducer, ... }  // Right here! We have a reducer!
    }),
  });
```

So what's going on?

### Reducer Registration Issue

LangGraph stores reducers in `schemaMetaRegistry`. When you call `withLangGraph()`, it does:

```typescript
  schemaMetaRegistry.set(zodSchemaObject, { reducer: yourReducer });
```

The critical thing here: the WeakMap is keyed by object identity. It's not checking if the schemas are structurally equivalent - it's checking if they're the exact same object in memory.

###  Problem 1: Duplicate packages

This is actually what's going on in issues 52 and 65, which covers up the true underlying issue (below)

When I personally encountered this bug, I ran:

```bash
npm ls @langchain/langgraph
```

And saw two different paths:

```bash
  deepagents → @langchain/langgraph@1.0.2_...3kxia2yhuhf7i35oisvqsvwple
  my-app  → @langchain/langgraph@1.0.2_...krfp3ij63gq4bje2wvj2d6skxu
```

Same version number, but pnpm created two separate copies because the dependency trees differed slightly. Each copy has its own schemaMetaRegistry WeakMap.

So deepagents registers the reducer in WeakMap A, but the app's graph looks it up in WeakMap B. WeakMap B is empty. No reducer found. Falls back to LastValue. 

**This causes parallel tool calls to fail, even though we're supposed to have a reducer**

### The fix for bug 1 (the bug everyone is reporting)

Add pnpm overrides to force a single copy (I found I had to pin all 3 of these to get a single, consistent `schemaMetaRegistry`. It's worth considering whether this singleton pattern in the core codebase might be refactored another way to avoid these kinds of issues.

```json
  "pnpm": {
    "overrides": {
      "@langchain/langgraph": "1.0.2",
      "@langchain/core": "^1.0.0",
      "zod": "3.25.76"
    }
  }
```

### Problem 2: A real bug in deepagents

With packages deduped, I now got a different error:

```bash
  Channel "files" already exists with a different type.
```

Now LangGraph sees the reducer, but it's complaining about duplicate channel definitions.

### The real culprit

Look at where FilesystemStateSchema was defined:

```typescript
  function createFilesystemMiddleware() {
    // This runs EVERY time the function is called
    const FilesystemStateSchema = z3.object({
      files: withLangGraph(...)
    });

    return createMiddleware({ stateSchema: FilesystemStateSchema });
  }
```

Every call to createFilesystemMiddleware() creates a brand new Zod object. And createDeepAgent calls it multiple times:

  1. Once for the parent agent's middleware
  2. Once for each subagent's defaultMiddleware

Each call registers a different schema object in the WeakMap. When LangGraph tries to compose the graph, it sees multiple files channels with different object identities and throws.

### The fix

Move the schema to module level so it's created only once. One WeakMap entry. No conflicts. Reducer works. Parallel tools work.

### TL;DR

Two bugs stacked on top of each other:

  1. pnpm created duplicate package copies → separate WeakMaps → reducer lookup failed → LastValue used instead (this is the bug people are seeing in #52  + #65 )
  2. Schema created inside function → new object each call → multiple registrations → "channel already exists" error

Both needed to be fixed, but one is an issue on the user side, and one we can fix for deepagents package itself.

### DIFF
--- a/.changeset/fair-hounds-make.md
+++ b/.changeset/fair-hounds-make.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+Fix 'Channel "files" already exists with a different type.' error due to different schema identity


ponderingdemocritus:
ship it! package is broken right now until this is merged

brettshollenberger:
Killer, thanks for the great work on the library!

Issue #75, opened 2025-12-05
[bug] Channel "files" already exists with a different type.


ponderingdemocritus:
Running example and get this. Assuming a compatibility issue with langgraph

```
168 |                   let channel;
169 |                   if (typeof val === "function") channel = val();
170 |                   else channel = val;
171 |                   if (this.channels[key] !== void 0) {
172 |                           if (this.channels[key] !== channel) {
173 |                                   if (channel.lc_graph_name !== "LastValue") throw new Error(`Channel "${key}" already exists with a different type.`);
                                                                ^
error: Channel "files" already exists with a different type.
      at _addSchema (/Users/os/Documents/code/shell/node_modules/@langchain/langgraph/dist/graph/state.js:173:59)
      at new StateGraph (/Users/os/Documents/code/shell/node_modules/@langchain/langgraph/dist/graph/state.js:150:8)
      at new ReactAgent (/Users/os/Documents/code/shell/node_modules/langchain/dist/agents/ReactAgent.js:47:20)
      at createAgent (/Users/os/Documents/code/shell/node_modules/langchain/dist/agents/index.js:9:9)
      at getSubagents (/Users/os/Documents/code/shell/node_modules/deepagents/dist/index.js:878:31)
      at createTaskTool (/Users/os/Documents/code/shell/node_modules/deepagents/dist/index.js:911:73)
      at createSubAgentMiddleware (/Users/os/Documents/code/shell/node_modules/deepagents/dist/index.js:947:11)
      at createDeepAgent (/Users/os/Documents/code/shell/node_modules/deepagents/dist/index.js:1891:3)
      at /Users/os/Documents/code/shell/apps/deep-agent/test.ts:200:22
```

kschluter:
Getting the same error. I tried the patch @janpawellek mentioned and it didn't seem to resolve it. Are there any releases that still work now? I tried installing all of them and either got this error or an error related to @langchain imports.

sinedied:
Same as @kschluter, I tried the #74 patch and still getting the error.

agntux:
I'm also getting this error after updating package versions. @sinedied - are you saying that you updated the @langchain/core package to 1.0.6 or something else? I've tried a few different combinations of versions for the deepagents package, langgraph, lanchain core, etc and can't find a working combination. 

Here are my current versions:
    "@langchain/core": "^1.1.4",
    "@langchain/langgraph": "^1.0.4",
[...]

PR #84, opened 2025-12-26
feat: support SystemMessage for systemPrompt parameter

### LINKED ISSUE #83: feat: support SystemMessage for systemPrompt parameter in createDeepAgent()
## Feature Request

Currently, `createDeepAgent()` only accepts a `string` for the `systemPrompt` parameter:

```typescript
systemPrompt?: string;
```

However, langchain's underlying `createAgent()` already supports both `string` and `SystemMessage`:

```typescript
systemPrompt?: string | SystemMessage;
```

## Problem

This creates friction for projects that model system prompts as structured `SystemMessage` objects. Users must either manually extract the string content or maintain separate conversion logic just to use DeepAgents.

## Proposed Solution

Expand the type signature to accept both strings and `SystemMessage` instances:

```typescript
systemPrompt?: string | SystemMessage;
```

If a `SystemMessage` is provided, the function should internally extract its content and use it following DeepAgents' existing prompt composition logic.

This maintains full backward compatibility with string inputs while enabling cleaner integration with higher-level prompt abstraction patterns.

ps06756:
## Summary

- Allow `createDeepAgent()` to accept `SystemMessage` objects in addition to strings for the `systemPrompt` parameter
- Aligns with the capability already available in langchain's `createAgent()`
- Enables cleaner integration for projects that model system prompts as structured `SystemMessage` objects

Closes #83

## Changes

1. Import `SystemMessage` from `@langchain/core/messages`
2. Update `systemPrompt` type in `CreateDeepAgentParams` interface to `string | SystemMessage`
3. Add logic to extract content from `SystemMessage` if provided
4. Add tests for both string and `SystemMessage` system prompts

## Test plan

- [x] Added unit tests for string systemPrompt
- [x] Added unit tests for SystemMessage systemPrompt
- [ ] Note: Existing test failures are due to a known upstream bug (Issue #75) unrelated to this PR

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -14,6 +14,7 @@ import type {
   BaseCheckpointSaver,
   BaseStore,
 } from "@langchain/langgraph-checkpoint";
+import { SystemMessage } from "@langchain/core/messages";
 
 import {
   createFilesystemMiddleware,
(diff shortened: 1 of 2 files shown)


src/agent.ts
```diff
@@ -106,9 +105,16 @@ export function createDeepAgent<
...
+  const customPromptContent = systemPrompt
+    ? systemPrompt instanceof SystemMessage
+      ? systemPrompt.content
+      : systemPrompt
+    : undefined;
+
   // Combine system prompt with base prompt like Python implementation
-  const finalSystemPrompt = systemPrompt
-    ? `${systemPrompt}\n\n${BASE_PROMPT}`
+  const finalSystemPrompt = customPromptContent
```
christian-bromann, line 116: I don't think we want to stringify a SystemMessage object as we would loose all features that comes with it. We made changes to `createAgent` which allows now `SystemMessage` as parameter

ps06756:
@christian-bromann It would have been a much better etiquette if you had added your commit here and merged the PR. 

This totally cuts away my contribution

PR #102, opened 2026-01-10
fix: resolve instanceof AIMessageChunk failing for Ollama provider

### LINKED ISSUE #64: instanceof AIMessageChunk Fails When Using Ollama Provider
### **Summary**

When streaming responses using the Ollama provider, the returned chunks **look like** `AIMessageChunk` objects, but `instanceof AIMessageChunk` returns `false`.
The exact same code **works as expected** when using OpenAI models.

This suggests that the Ollama integration is constructing message chunks using a *different module instance* of `AIMessageChunk`, causing the prototype chain to mismatch.

---

### **OpenAI (works)**

```
Raw chunk: AIMessageChunk {
  id: "chatcmpl-xxx",
  content: "kan",
  response_metadata: { model_provider: "openai", usage: {} },
  ...
}

is AIMessageChunk: true
constructor: AIMessageChunk
type: ai
```

### **Ollama (fails)**

```
Raw chunk: AIMessageChunk {
  id: "run-xxx",
  content: "al",
  response_metadata: {},
  ...
}

is AIMessageChunk: false
constructor: AIMessageChunk
type: ai
```

Notice both objects expose the same fields and the same constructor name, but `instanceof` returns `false`.

---

## **Diagnosis**

This typically happens when a class is constructed from **a different copy of the module**.
Most likely:

* The Ollama DeepAgent integration imports `AIMessageChunk` from a bundled/isolated instance of `@langchain/core`
* The application imports `AIMessageChunk` from a different instance
  → Therefore the prototype comparison inside `instanceof` fails.

This does **not** happen with OpenAI provider, indicating consistent import paths on that side.

---

## **Expected Behavior**

`instanceof AIMessageChunk` should return `true` for all providers that generate `AIMessageChunk` objects, including Ollama.

---

## **Why This Causes Problems**

Libraries and downstream apps frequently use:

```ts
if (chunk instanceof AIMessageChunk) {
    // handle AI message chunks
}
```

Because Ollama chunks fail this check, they require special-case logic that breaks the uniformity of message streaming.

---

## **Possible Fixes**

1. Ensure the Ollama provider constructs chunks using the same import path:

   ```ts
   import { AIMessageChunk } from "@langchain/core/messages";
   ```

2. Avoid bundling or including a second copy of `@langchain/core`.

3. Re-export and use a single instance from deepagentjs to guarantee consistency.

---

## **Additional Notes**

Despite failing `instanceof`, the objects are valid message chunks and `.type === "ai"` works correctly.
Normalizing manually also fixes the issue:

```ts
new AIMessageChunk({ ...chunk });
```

But ideally, this should not be required.

VedantMadane:
Closes #64

## Problem
When streaming responses using the Ollama provider, chunks look like AIMessageChunk objects but instanceof AIMessageChunk returns false. This happens because @langchain/core was being bundled, creating a duplicate module instance with a different prototype chain.

## Solution
1. Externalize @langchain/core in tsdown bundler config to prevent duplicate module instances
2. Replace internal instanceof ToolMessage checks with the more robust ToolMessage.isInstance() pattern

## Testing
- Added integration test that verifies Ollama-streamed chunks pass both instanceof AIMessageChunk and AIMessageChunk.isInstance()
- All 178 existing unit tests pass

### DIFF
--- a/libs/deepagents/src/ollama-instanceof.int.test.ts
+++ b/libs/deepagents/src/ollama-instanceof.int.test.ts
@@ -1,0 +1,82 @@
+/**
+ * Integration test for instanceof AIMessageChunk with Ollama provider.
+ *
+ * This test verifies that the fix for issue #64 works correctly:
+ * - AIMessageChunk instances from Ollama streaming should pass `instanceof AIMessageChunk`
+ * - This was previously failing because of duplicate @langchain/core module instances
+ *
+ * Prerequisites:
+ * - Ollama must be running locally (`ollama serve`)
+ * - A model must be available (e.g., `ollama pull tinyllama`)
+ */
+
+import { describe, it, expect } from "vitest";
+import { ChatOllama } from "@langchain/ollama";
+import { AIMessageChunk } from "@langchain/core/messages";
+
+describe("Ollama instanceof AIMessageChunk", () => {
+  it.concurrent(
+    "should return true for instanceof AIMessageChunk when streaming",
+    { timeout: 60_000 },
+    async () => {
+      const model = new ChatOllama({
+        model: "tinyllama",
+        baseUrl: "http://localhost:11434",
+      });
+
+      const chunks: AIMessageChunk[] = [];
+      let instanceofPassCount = 0;
+      let isInstancePassCount = 0;
+
+      // Stream a simple response
+      const stream = await model.stream("Say 'hello' in one word.");
+
+      for await (const chunk of stream) {
+        chunks.push(chunk);
+
+        // Test both methods
+        if (chunk instanceof AIMessageChunk) {
+          instanceofPassCount++;
+        }
+        if (AIMessageChunk.isInstance(chunk)) {
+          isInstancePassCount++;
+        }
+
+        // Debug info available in verbose mode:
+        // content, instanceofPass, isInstancePass, constructorName
+      }
+
+      // We should have received at least one chunk
+      expect(chunks.length).toBeGreaterThan(0);
+
+      // All chunks should pass AIMessageChunk.isInstance() - this always works
+      expect(isInstancePassCount).toBe(chunks.length);
+
+      // After the fix: All chunks should ALSO pass instanceof AIMessageChunk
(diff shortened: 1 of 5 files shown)


libs/deepagents/tsdown.config.ts
```diff
@@ -9,6 +9,10 @@ export default defineConfig([
     sourcemap: true,
     outDir: "dist",
     outExtensions: () => ({ js: ".js" }),
+    // Externalize @langchain/core to prevent bundling duplicate copies.
+    // This fixes instanceof checks (e.g. AIMessageChunk) failing when
+    // providers like Ollama construct chunks from a different module instance.
+    external: [/^@langchain\/core/],
```
christian-bromann, line 15: This should be automatically set since `@langchain/core` is part of dependency list. Let's remove it.

libs/deepagents/tsdown.config.ts
```diff
@@ -18,5 +22,7 @@ export default defineConfig([
     sourcemap: true,
     outDir: "dist",
     outExtensions: () => ({ js: ".cjs" }),
+    // Externalize @langchain/core to prevent bundling duplicate copies.
+    external: [/^@langchain\/core/],
```
christian-bromann, line 26: same

libs/deepagents/src/ollama-instanceof.int.test.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Prerequisites:
+ * - Ollama must be running locally (`ollama serve`)
+ * - A model must be available (e.g., `ollama pull tinyllama`)
+ */
+
+import { describe, it, expect } from "vitest";
+import { ChatOllama } from "@langchain/ollama";
+import { AIMessageChunk } from "@langchain/core/messages";
+
+describe("Ollama instanceof AIMessageChunk", () => {
```
christian-bromann, line 17: I don't think we need this extra integration test just for this.

christian-bromann (changes requested):
Can we add an eslint rule and disallow `instanceof` comparisons?

PR #111, opened 2026-01-12
Ensure register files reducer in Langgraph schemaMetaRegistry

brettshollenberger:
## Bug

Related to https://github.com/langchain-ai/deepagentsjs/issues/52 and https://github.com/langchain-ai/deepagentsjs/issues/65 and https://github.com/langchain-ai/deepagentsjs/issues/75

## Expectation

FS middleware should have custom reducer. Instead, it receives `LastValue` reducer. This means backends which return files to state receive the error: `LastValue can only receive one value per step`

## Root Cause

Zod 4's .meta() stores metadata in Zod's internal globalRegistry, but LangGraph reads from its own separate schemaMetaRegistry. The reducer was being registered in the wrong place, so LangGraph couldn't find it and fell back to LastValue (single-value-only channel) instead of BinaryOperatorAggregate (merge reducer).
 
@christian-bromann since you've been looking at stuff related to this

### DIFF
--- a/libs/deepagents/src/middleware/fs.ts
+++ b/libs/deepagents/src/middleware/fs.ts
@@ -13,6 +13,7 @@ import {
   type AgentMiddleware as _AgentMiddleware,
 } from "langchain";
 import { Command, isCommand, getCurrentTaskInput } from "@langchain/langgraph";
+import { withLangGraph } from "@langchain/langgraph/zod";
 import { z } from "zod/v4";
 import type {
   BackendProtocol,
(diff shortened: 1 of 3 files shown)


libs/deepagents/src/middleware/fs.ts
```diff
@@ -77,15 +78,12 @@ function fileDataReducer(
...
-  files: z
-    .record(z.string(), FileDataSchema)
-    .default({})
-    .meta({
-      reducer: {
-        fn: fileDataReducer,
-        schema: z.record(z.string(), FileDataSchema.nullable()),
-      },
-    }),
+  files: withLangGraph(z.record(z.string(), FileDataSchema).default({}), {
```
christian-bromann, line 81: I think given that `withLangGraph` particularly relies on Zod v3 I am not sure if this is the best approach here.

brettshollenberger: Interesting, following along! This fix would be great to see, have got some workarounds now to merge files back to state after the agent runs 😵

daniel-watts-ai: @brettshollenberger What is your workaround? We're also seeing this issue in `1.4.1`.

hntrl:
Closing since this was fixed with #145

PR #157, opened 2026-01-31
Fix memory loading for StateBackend invocations

maahir30:
## Summary
- Fix memory loading when using the `StateBackend` by lazily loading memory inside `wrapModelCall` once `files` are actually present in request state.
- Cache loaded memory so the very first model call sees the memory section in the system prompt.
- Update memory middleware tests to await async `wrapModelCall` and add a StateBackend‑specific regression test.

## Problem & reasoning
When using the `StateBackend`, memory files like `/AGENTS.md` are provided **at invoke time** via the `files` input. The memory middleware tries to load memory in `beforeAgent`, but **at that moment the input files are not yet merged into state**, so the backend can’t read them and `memoryContents` becomes `{}`. Later in the run, tools can read `/AGENTS.md` because the files are merged by then, but the system prompt already missed the memory content.

This issue does **not** appear with the StoreBackend because the file already exists in the persistent store before the run starts.

We considered “moving `beforeAgent` later,” but that would require changing LangGraph’s invocation pipeline. Instead, we implemented the smallest fix: **lazy‑load memory right before the model call**, when request state already includes the input files. We cache the loaded memory so it’s reused for subsequent calls in the same run.

## Fix (conceptual)
- Keep `beforeAgent` as‑is for backends that already have files.
- If memory is still empty at `wrapModelCall` time and files are present, load memory then.
- Inject the loaded memory into the system prompt for the very first model call.

### DIFF
--- a/libs/deepagents/src/middleware/memory.test.ts
+++ b/libs/deepagents/src/middleware/memory.test.ts
@@ -1,5 +1,8 @@
 import { describe, it, expect, vi } from "vitest";
 import { createMemoryMiddleware } from "./memory.js";
+import { StateBackend } from "../backends/state.js";
+import { createFileData } from "../backends/utils.js";
+import type { BaseStore } from "@langchain/langgraph-checkpoint";
 import type {
   BackendProtocol,
   FileDownloadResponse,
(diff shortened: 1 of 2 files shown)


tianshangwuyun:
Skills seem to have the same issue, have they all been handled uniformly

PR #160, opened 2026-02-01
fix: missing files in skillsMiddleware schema

maahir30:
## Summary

- Add `files` field to `SkillsStateSchema` so `StateBackend` can access files
- Export `FileDataSchema` from `fs.ts` for reuse across middleware

## Problem

`SkillsMiddleware` fails to load skill files when using `StateBackend` because `SkillsStateSchema` doesn't declare the `files` field. LangGraph filters state passed to middleware based on their declared schema, so `beforeAgent` receives state without `files`, causing `StateBackend` operations to fail.

## Solution

Add the `files` field to `SkillsStateSchema` using the same `ReducedValue` pattern as `FilesystemStateSchema`:

```typescript
files: new ReducedValue(
  z.record(z.string(), FileDataSchema).default(() => ({})),
  {
    inputSchema: z.record(z.string(), FileDataSchema.nullable()).optional(),
    reducer: fileDataReducer,
  },
),
```

### DIFF
--- a/.changeset/fix-skills-middleware.md
+++ b/.changeset/fix-skills-middleware.md
@@ -1,0 +1,9 @@
+---
+"deepagents": patch
+---
+
+fix(skills): properly restore skills from StateBackend checkpoint
+
+- Add `files` channel to `SkillsStateSchema` for StateBackend integration
+- Fix skills restoration check to require non-empty array instead of just non-null
+- Export `FileDataSchema` from fs middleware for reuse


christian-bromann (commented):
Awesome! Can we add a unit or integration test to verify we don't regress here?

libs/deepagents/src/middleware/skills.int.test.ts
```diff
@@ -0,0 +1,266 @@
...
+      const content = lastMessage.content.toString().toLowerCase();
+
+      // Both skills from different sources should be loaded
+      expect(content).toContain("user-skill");
+      expect(content).toContain("project-skill");
+    },
+  );
+
+  it.concurrent(
+    "should handle empty skills directory gracefully",
```
christian-bromann: This can be a unit test

libs/deepagents/src/middleware/skills.int.test.ts
```diff
@@ -0,0 +1,266 @@
...
+        } as any,
+        { ...config, recursionLimit: 50 },
+      );
+
+      const lastMessage = response.messages[response.messages.length - 1];
+      const content = lastMessage.content.toString().toLowerCase();
+
+      // The description should be visible to the model
+      expect(content).toContain("quantum");
+      expect(content).toContain("entanglement");
```
christian-bromann: This doesn't seem like we verify the system prompt. I suggest move this into a unit test, use a [mocked chat instance](https://github.com/langchain-ai/langchainjs/blob/2d744995b8194302308c4f61e8690f25df2b7c62/libs/langchain/src/agents/tests/utils.ts#L116-L280) and verify that the first message contains the skills content.

libs/deepagents/src/middleware/skills.int.test.ts
```diff
@@ -0,0 +1,266 @@
...
+        configurable: {
+          thread_id: `test-skills-${Date.now()}`,
+        },
+      };
+
+      const response = await agent.invoke(
+        {
+          messages: [
+            new HumanMessage(
+              "Based on the skills system shown in your prompt, what is the name of the available skill? Just say the skill name, nothing else.",
```
christian-bromann: > Just say the skill name, nothing else.

Maybe it is better to work with structured output here?

libs/deepagents/src/middleware/skills.test.ts
```diff
@@ -672,3 +678,258 @@ describe("skillsMetadataReducer", () => {
     });
   });
 });
+
+/**
+ * StateBackend integration tests.
+ *
+ * These tests verify that skills are properly loaded from state.files and
+ * injected into the system prompt when using createDeepAgent with StateBackend.
+ */
```
christian-bromann, line 690:
```suggestion
```

libs/deepagents/src/middleware/skills.test.ts
```diff
@@ -672,3 +678,258 @@ describe("skillsMetadataReducer", () => {
...
+    const model = new FakeListChatModel({
+      responses: ["Done"],
+    });
+
+    const checkpointer = new MemorySaver();
+    const agent = createDeepAgent({
+      model: model as any,
+      skills: ["/skills/"],
+      checkpointer,
+      middleware: [createCaptureMiddleware(capture)],
```
christian-bromann: Let's not use a middleware for this, instead just mock the `invoke`

libs/deepagents/src/middleware/skills.test.ts
```diff
@@ -672,3 +677,266 @@ describe("skillsMetadataReducer", () => {
...
+   * Helper to extract system prompt content from model invoke spy.
+   * The system message can have content as string or array of content blocks.
+   */
+  function getSystemPromptFromSpy(
+    invokeSpy: ReturnType<typeof vi.spyOn>,
+  ): string {
+    const lastCall = invokeSpy.mock.calls[invokeSpy.mock.calls.length - 1];
+    const messages = lastCall?.[0] as BaseMessage[] | undefined;
+    if (!messages) return "";
+    const systemMessage = messages.find((m) => m.getType() === "system");
```
christian-bromann:
```suggestion
    const systemMessage = messages.find(SystemMessage.isInstance);
```

libs/deepagents/src/middleware/skills.test.ts
```diff
@@ -672,3 +677,266 @@ describe("skillsMetadataReducer", () => {
...
+    if (!systemMessage) return "";
+
+    const content = systemMessage.content;
+    if (typeof content === "string") return content;
+    if (Array.isArray(content)) {
+      return content
+        .map((block) => (typeof block === "string" ? block : block.text || ""))
+        .join("");
+    }
+    return "";
```
christian-bromann: `SystemMessage.text` is a getter that [does exactly](https://github.com/langchain-ai/langchainjs/blob/acd8ca4f43ff84b6fe04445b451ea20e55b96181/libs/langchain-core/src/messages/base.ts#L296-L309) what you do here 😉 

```suggestion
    return systemMessage.text;
```

christian-bromann (commented):
Awesome, two last comments

christian-bromann (approved):
🙌

PR #165, opened 2026-02-02
feat: add `SandboxProvider` abstraction

christian-bromann:
Ports the `SandboxProvider` protocol from Python PR #900 to TypeScript. This adds a standardized interface for sandbox lifecycle management (list, create, delete), separate from sandbox execution.

### Summary

- Add `SandboxProvider<MetadataT>` interface for third-party sandbox integrations
- Add `SandboxInfo` and `SandboxListResponse` types for paginated sandbox listings
- Implement `VercelSandboxProvider` in `@langchain/vercel-sandbox`
- Implement `VfsSandboxProvider` in `@langchain/vfs-sandbox`
- Add comprehensive unit tests for protocol compliance

### Motivation

The existing `Sandbox` classes mix execution (run commands, file ops) with lifecycle management (create, list, delete). This separation allows:

1. **CLI tools** to list/manage user's sandboxes across providers
2. **Centralized resource management** for cleanup
3. **Consistent API** across different sandbox backends

### Usage

```typescript
import { VercelSandboxProvider } from "@langchain/vercel-sandbox";

const provider = new VercelSandboxProvider({ runtime: "node24" });

// Create a new sandbox
const sandbox = await provider.getOrCreate();

// Execute commands using the sandbox
const result = await sandbox.execute("node --version");

// List sandboxes (if supported by provider)
const { items } = await provider.list();

// Delete when done (idempotent)
await provider.delete({ sandboxId: sandbox.id });
```

### DIFF
--- a/.changeset/real-swans-crash.md
+++ b/.changeset/real-swans-crash.md
@@ -1,5 +1,5 @@
 ---
-"deepagents": patch
+"deepagents": minor
 ---
 
 feat: add SandboxProvider abstraction


libs/deepagents/src/backends/index.ts
```diff
@@ -21,6 +21,13 @@ export type {
   FileUploadResponse,
   SandboxBackendProtocol,
   MaybePromise,
+  // Sandbox provider types
+  SandboxInfo,
```
eyurtsev, line 25: This feels to me like a large collection of primitives that are specific to sandboxes that are being exposed in a backends top level space. Is there a reasonable way to encapsulate this inside sandboxes namespace?

christian-bromann: In JS we tend to export all primitives through the main namespace. Only if certain primitives are environment specific we create separate namespaces, e.g. `import * from "deepagents/browser-sdk". Do we expect a lot of collisions here?

Note: it doesn't matter where the type is exported. They often end up being [re-exported in other files](https://github.com/langchain-ai/deepagentsjs/blob/bc4bb537ac42891930a53e4754409b3851c97dfc/libs/deepagents/src/index.ts#L72-L100), e.g. and the [package.json](https://github.com/langchain-ai/deepagentsjs/blob/bc4bb537ac42891930a53e4754409b3851c97dfc/libs/deepagents/package.json#L67-L79) is the file that decides how exports look like.

PR #180, opened 2026-02-03
fix(skills): isolate skillsMetadata between parent agent and subagents

maahir30:
This PR exclude skillsMetadata from subagent state updates to prevent concurrent update errors

**1. Added `skillsMetadata` to `EXCLUDED_STATE_KEYS`**

Skills are loaded per-agent by the skills middleware, so there's no need to propagate `skillsMetadata` between parent and child agents:
- Subagents don't receive parent's `skillsMetadata` 
- Subagents don't return their `skillsMetadata` to parent

This prevents concurrent update conflicts when parallel subagents complete simultaneously.

**2. Fixed skills middleware to only apply to general-purpose subagent**

Skills middleware is now added via the `generalPurposeMiddleware` parameter instead of `defaultMiddleware`:
- General-purpose subagent: Gets skills middleware (inherits from parent config)
- Custom subagents: Do NOT get skills middleware (define their own middleware explicitly)

Added unit tests that verify:
- Custom subagents do NOT inherit skills middleware
- `skillsMetadata` does NOT bubble up from subagent to parent  
- General-purpose subagent DOES inherit skills middleware

### DIFF
--- a/libs/deepagents/src/middleware/subagents.ts
+++ b/libs/deepagents/src/middleware/subagents.ts
@@ -28,10 +28,15 @@ const DEFAULT_SUBAGENT_PROMPT =
 // 1. The messages key is handled explicitly to ensure only the final message is included
 // 2. The todos and structuredResponse keys are excluded as they do not have a defined reducer
 //    and no clear meaning for returning them from a subagent to the main agent.
+// 3. The skillsMetadata key is excluded because skills are loaded per-agent
+//    -> Subagents don't receive the parent's skillsMetadata.
+//    -> Subagents don't return their skillsMetadata to the parent.
+
 const EXCLUDED_STATE_KEYS = [
   "messages",
   "todos",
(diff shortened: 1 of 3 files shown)


libs/deepagents/src/middleware/subagents.ts
```diff
@@ -435,6 +449,8 @@ export interface SubAgentMiddlewareOptions {
   defaultTools?: StructuredTool[];
   /** Default middleware to apply to all subagents */
   defaultMiddleware?: AgentMiddleware[] | null;
+  /** Additional middleware to apply only to the general-purpose subagent (e.g., SkillsMiddleware). Custom subagents do not inherit this middleware. */
+  generalPurposeMiddleware?: AgentMiddleware[] | null;
```
hntrl, line 453: not totally sold we want to separate middleware out like this, seems like awkward API design

PR #181, opened 2026-02-03
refactor: Add shared filesValue state primitive

maahir30:
### Add shared `filesValue` state primitive

Introduces `filesValue` - a reusable `ReducedValue` for file state management, similar to LangGraph's `messagesValue`.

Replaces duplicated inline `ReducedValue` definitions in `memory.ts` and `skills.ts` with a shared value exported from new `values.ts` module.

Users can now import `filesValue` directly for custom state schemas.


christian-bromann (approved):
LGTM 👍

PR #189, opened 2026-02-04
fix(summarization): resolve fraction trigger bug by using model profile for maxInputTokens

### LINKED ISSUE #185: `createSummarizationMiddleware` logic error
## Description

It seems that there is a logic issue in the `createSummarizationMiddleware` middleware in the main branch.

path: libs/deepagents/src/middleware/summarization.ts

```ts
    if (truncateTrigger.type === "fraction" && maxInputTokens) {
      const threshold = Math.floor(maxInputTokens * truncateTrigger.value);
      return totalTokens >= threshold;
    }
```

I pulled down the main branch to test it and found that the fraction parameter is always invalid because maxInputTokens is always undefined.

christian-bromann:
Fixes a bug where the `fraction` trigger type in `createSummarizationMiddleware` was always ineffective because `maxInputTokens` was never passed to the internal functions.

**Before:** Fraction-based triggers (e.g., `{ type: "fraction", value: 0.85 }`) never fired because `maxInputTokens` was always `undefined`.

**After:** The middleware automatically retrieves `maxInputTokens` from the model's profile, aligning with the Python implementation.

## Changes

- **Use `initChatModel`**: Replaced `ChatOpenAI` with `initChatModel` from `langchain/chat_models/universal` to resolve string model names. This supports any provider (OpenAI, Anthropic, Google, etc.)
- **Automatic `maxInputTokens` resolution**: Get `maxInputTokens` from `model.profile.maxInputTokens` (similar to Python's `_get_profile_limits`)
- **Removed explicit `maxInputTokens` option**: No longer needed since it's derived from the model profile
- **Model caching**: Cache the resolved model to avoid repeated `initChatModel` calls on each `beforeModel` invocation
- **Updated tests**: Added tests for fraction triggers and updated mocks to use `initChatModel`

## Usage

```typescript
// Works with any model provider - maxInputTokens comes from profile automatically
const middleware = createSummarizationMiddleware({
  model: "anthropic:claude-opus-4-0", // or "openai:gpt-4", etc.
  backend,
  trigger: { type: "fraction", value: 0.85 },
  keep: { type: "fraction", value: 0.10 },
});

// Or with model instance
const model = new ChatAnthropic({ model: "claude-opus-4-0" });
const middleware = createSummarizationMiddleware({
  model,
  backend,
  trigger: { type: "fraction", value: 0.85 },
});
```

fixes #185

### DIFF
--- a/libs/deepagents/src/middleware/summarization.ts
+++ b/libs/deepagents/src/middleware/summarization.ts
@@ -287,9 +287,7 @@ export function createSummarizationMiddleware(
    * Similar to Python's _get_profile_limits.
    */
   function getMaxInputTokens(resolvedModel: BaseChatModel): number | undefined {
-    const profile = (
-      resolvedModel as BaseChatModel & { profile?: { maxInputTokens?: number } }
-    ).profile;
+    const profile = resolvedModel.profile;
     if (
       profile &&
       typeof profile === "object" &&


libs/deepagents/src/middleware/summarization.ts
```diff
@@ -259,14 +259,46 @@ export function createSummarizationMiddleware(
...
+  }
+
+  /**
+   * Get the max input tokens from the resolved model's profile.
+   * Similar to Python's _get_profile_limits.
+   */
+  function getMaxInputTokens(resolvedModel: BaseChatModel): number | undefined {
+    const profile = (
+      resolvedModel as BaseChatModel & { profile?: { maxInputTokens?: number } }
+    ).profile;
```
hntrl: you shouldn't need to do type gymnastics, this is available on the base class: https://github.com/langchain-ai/langchainjs/blob/4bb9ea0bf821d9563e4e890997ae31d9d7b403c5/libs/langchain-core/src/language_models/base.ts#L614-L616

hntrl (approved):

PR #193, opened 2026-02-06
feat: allow filesystemOptions to be passed to built-in middleware

chopraj:
Allow fileSystemOption to be passed into built-in file system middleware. Helps for those trying to model things other than files into a filesystem structure where the LLM might need additional guidance


christian-bromann:
Can we run `pnpm format`?

PR #201, opened 2026-02-07
test(sandbox): enhance standard tests for sandboxes

christian-bromann:
This PR enhances or standard tests for sandboxes.

### DIFF
--- a/internal/standard-tests/src/tests/initial-files.ts
+++ b/internal/standard-tests/src/tests/initial-files.ts
@@ -157,13 +157,13 @@ export function registerInitialFilesTests<T extends SandboxInstance>(
           config.createSandbox({
             initialFiles: {
               [scriptPath]:
-                '#!/bin/bash\necho "Hello from initialFiles script"',
+                '#!/bin/sh\necho "Hello from initialFiles script"',
             },
           }),
         );
 
         try {
-          const result = await tmp.execute(`bash ${scriptPath}`);
+          const result = await tmp.execute(`sh ${scriptPath}`);
           expect(result.exitCode).toBe(0);
           expect(result.output.trim()).toBe(
             "Hello from initialFiles script",


everywheredennise-oss:
🤖 **Devin AI** is starting automated code review...

libs/deepagents/src/backends/sandbox.ts
```diff
@@ -129,31 +136,70 @@ function parseStatLine(
...
+ * (supported by both GNU coreutils and BusyBox) for size and mtime.
+ * printf handles tab-delimited output formatting.
+ */
+const STAT_C_SCRIPT =
+  "for f; do " +
+  'if [ -d "$f" ]; then t=d; elif [ -L "$f" ]; then t=l; else t=f; fi; ' +
+  'sz=$(stat -c %s "$f" 2>/dev/null) || continue; ' +
+  'mt=$(stat -c %Y "$f" 2>/dev/null) || continue; ' +
+  'printf "%s\\t%s\\t%s\\t%s\\n" "$sz" "$mt" "$t" "$f"; ' +
+  "done";
```
hntrl, line 159: at some point we should set up a matrix test + benchmarks for these

libs/providers/deno/src/sandbox.ts
```diff
@@ -511,86 +513,6 @@ awk -v offset=${safeOffset} -v limit=${safeLimit} '
...
-
-    const editResult = await this.execute(awkCommand);
-
-    if (editResult.exitCode !== 0) {
-      return { error: `Unknown error editing file '${filePath}'` };
-    }
-
-    return { path: filePath, filesUpdate: null, occurrences: count };
-  }
-
```
hntrl, line 593: how come we're removing these but not read/write?

internal/standard-tests/package.json
hntrl, line 1: nit: can we move this standard-tests package to `libs/` to match convention?

hntrl: We should also think about standard model tests, backend tests, etc. (doesn't need to be apart of this PR)

christian-bromann: > nit: can we move this standard-tests package to libs/ to match convention?

I think we should rather "fix" this in langchainjs, since we don't publish this package, it shouldn't be among the packages we publish.

libs/providers/deno/src/sandbox.ts
christian-bromann: removed read/write too! Good catch 👌

PR #205, opened 2026-02-08
fix(deepagents): shell commands for ls and glob tools output a string that cannot be parsed

alvedder:
## Problem description

The new POSIX shell commands for `ls` and `glob` introduced in https://github.com/langchain-ai/deepagentsjs/pull/88 have `\\t` in their patterns.
This produces commands like that:
```bash
find '/workspace' -not -path '/workspace' -exec stat -c '%s\\t%Y\\t%F\\t%n' {} + 2>/dev/null || true
```
and they in turn produce output like this:
```
632\\t1770575029\\tregular file\\t/workspace/dragon_ascii_art.txt\n25\\t1770330477\\tregular file\\t/workspace/ascii_art_cat.txt\n
```
then this output is split into lines and passed to [`parseStatLine`](https://github.com/langchain-ai/deepagentsjs/blob/main/libs/deepagents/src/backends/sandbox.ts#L110-L135) and at [line 113](https://github.com/langchain-ai/deepagentsjs/blob/main/libs/deepagents/src/backends/sandbox.ts#L113) it essentially always returns `-1`:

```js
"632\\t1770575029\\tregular file\\t/workspace/dragon_ascii_art.txt".indexOf("\t") // returns -1
```
Because `\\t` is interpreted as `\ + t`, not `\ + \t`:
```js
'\\' + 't' === '\\t' // returns true
```

## Fix

Drop extra `\` in the patterns

### DIFF
--- a/libs/deepagents/src/backends/sandbox.ts
+++ b/libs/deepagents/src/backends/sandbox.ts
@@ -142,7 +142,7 @@ function parseStatLine(
  */
 function buildLsCommand(dirPath: string): string {
   const quotedPath = shellQuote(dirPath);
-  return `find ${quotedPath} -maxdepth 1 -not -path ${quotedPath} -exec stat -c '%s\\t%Y\\t%F\\t%n' {} + 2>/dev/null || true`;
+  return `find ${quotedPath} -maxdepth 1 -not -path ${quotedPath} -exec stat -c "%s\t%Y\t%F\t%n" {} + 2>/dev/null || true`;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


christian-bromann:
@alvedder thanks for the contribution, can you double check if issue is still appearing in #201 ?

christian-bromann:
We have published some improvements on cross platform compatibility for sandboxes in v1.7.3. Would you mind to double check whether this also resolves the issue you experienced?

PR #208, opened 2026-02-09
fix(skills): use systemMessage.concat() instead of systemPrompt string

antonnak:
## Problem

`SkillsMiddleware.wrapModelCall` reads from `request.systemPrompt` (a string) and writes back to `systemPrompt`. When a `SystemMessage` with structured content blocks is passed to `createDeepAgent` (supported since PR #104), this middleware:

1. Reads `request.systemPrompt` which is the `.text` getter output (flattened string)
2. Appends skills section via string concatenation
3. Sets `systemPrompt: newString` on the handler request

This triggers `AgentNode`'s `hasSystemPromptChanged` branch (AgentNode.js:219-229), which recreates the `SystemMessage` from scratch as a single `{type: "text", text: "..."}` block — **destroying all existing content block metadata** including `cache_control` annotations for Anthropic prompt caching.

## Root Cause

`SkillsMiddleware` is the **only** middleware in deepagents that uses the string-based `request.systemPrompt` API. Both `FilesystemMiddleware` (fs.ts:973) and `SubAgentMiddleware` (subagents.ts:1390) correctly use `request.systemMessage.concat()`, which preserves content block structure.

## Fix

Replace string-based `systemPrompt` manipulation with `request.systemMessage.concat()`:

```diff
- const currentSystemPrompt = request.systemPrompt || "";
- const newSystemPrompt = currentSystemPrompt
-   ? `${currentSystemPrompt}\n\n${skillsSection}`
-   : skillsSection;
- return handler({ ...request, systemPrompt: newSystemPrompt });
+ const newSystemMessage = request.systemMessage.concat(skillsSection);
+ return handler({ ...request, systemMessage: newSystemMessage });

### DIFF
--- a/libs/deepagents/src/skills/index.int.test.ts
+++ b/libs/deepagents/src/skills/index.int.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, beforeEach, afterEach } from "vitest";
 import fs from "node:fs";
 import path from "node:path";
 import os from "node:os";
+import { SystemMessage } from "@langchain/core/messages";
 import { createSettings } from "../config.js";
 import { listSkills } from "./loader.js";
 import { createSkillsMiddleware } from "../middleware/skills.js";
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
Thank you 👍

PR #209, opened 2026-02-10
feat(deepagents): add SDK version metadata to agent traces

hntrl:
## Summary

Add SDK version metadata to all agent runs created via `createDeepAgent`. Every trace sent to LangSmith now includes `sdk: "deepagents-js"` and `sdk_version` in the run metadata, making it easy to identify and filter deep agent runs.

## Changes

- **`libs/deepagents/src/agent.ts`** — Declare a `__SDK_VERSION__` build-time constant and attach `{ sdk: "deepagents-js", sdk_version }` to the `.withConfig()` metadata on the compiled agent.
- **`libs/deepagents/tsdown.config.ts`** — Read the package version from `package.json` at build time and inject it via tsdown's `define` option, replacing `__SDK_VERSION__` with the literal version string in the bundle.
- **`libs/deepagents/vitest.config.ts`** — Mirror the same `define` so that `__SDK_VERSION__` resolves during tests.


libs/deepagents/tsdown.config.ts
```diff
@@ -1,8 +1,13 @@
 import { defineConfig } from "tsdown";
+import { readFileSync } from "node:fs";
```
christian-bromann:
```suggestion
import { version } from "./package.json"  with { type: "json" };
```

libs/deepagents/vitest.config.ts
```diff
@@ -1,4 +1,5 @@
 import path from "node:path";
+import { readFileSync } from "node:fs";
```
christian-bromann:
```suggestion
import { version } from "./package.json"  with { type: "json" };
```

libs/deepagents/tsdown.config.ts
```diff
@@ -1,8 +1,13 @@
 import { defineConfig } from "tsdown";
+import { version } from "./package.json"  with { type: "json" };
 
 // Mark all node_modules as external since this is a library
 const external = [/^[^./]/];
 
+const { version } = JSON.parse(
+  readFileSync(new URL("./package.json", import.meta.url), "utf-8"),
+);
+
```
christian-bromann, line 10:
```suggestion
```

PR #216, opened 2026-02-10
feat(deno): better support for sandbox volumes

christian-bromann:
This patch updates the DenoSandbox options (in a backwards compatible way) to allow users attach volumes to the sandbox.

### DIFF
--- a/libs/providers/deno/src/auth.test.ts
+++ b/libs/providers/deno/src/auth.test.ts
@@ -73,9 +73,7 @@ describe("getAuthToken", () => {
         getAuthToken();
       } catch (error) {
         expect(error).toBeInstanceOf(DenoSandboxError);
-        expect((error as DenoSandboxError).code).toBe(
-          "AUTHENTICATION_FAILED",
-        );
+        expect((error as DenoSandboxError).code).toBe("AUTHENTICATION_FAILED");
       }
     });
 


libs/providers/deno/src/sandbox.ts
```diff
@@ -541,25 +549,49 @@ export class DenoSandbox extends BaseSandbox {
...
-      throw new DenoSandboxError(
-        "Failed to authenticate with Deno Deploy. Check your token configuration.",
-        "AUTHENTICATION_FAILED",
-        error instanceof Error ? error : undefined,
-      );
+    // Resolve authentication: top-level `token` takes precedence over deprecated `auth.token`
+    const resolvedToken = options?.token ?? options?.auth?.token ?? undefined;
+
+    if (resolvedToken) {
+      process.env.DENO_DEPLOY_TOKEN = resolvedToken;
```
hntrl: why do we need to set process.env? if we're always passing it through sdk options surely we dont need to do this?

libs/providers/deno/src/types.ts
```diff
@@ -109,6 +169,161 @@ export interface DenoSandboxOptions {
...
+   *
+   * The key is the mount path inside the sandbox, and the value is the
+   * volume ID or slug.
+   *
+   * @example
+   * ```typescript
+   * const options: DenoSandboxOptions = {
+   *   volumes: {
+   *     "/data/volume1": "volume-slug-or-id-1",
+   *   },
```
hntrl, line 244: they let you do this?? 🐐 sandbox ngl

libs/providers/deno/src/sandbox.ts
```diff
@@ -171,39 +166,52 @@ export class DenoSandbox extends BaseSandbox {
...
+        credentials = getAuthCredentials(this.#options.auth);
+      } catch (error) {
+        throw new DenoSandboxError(
+          "Failed to authenticate with Deno Deploy. Check your token configuration.",
+          "AUTHENTICATION_FAILED",
+          error instanceof Error ? error : undefined,
+        );
       }
+      process.env.DENO_DEPLOY_TOKEN = credentials.token;
+    }
```
hntrl: nit: is it possible to have this setting/ error throwing logic inside of getAuthCredentials? Thats how we do it in Modal I believe

PR #218, opened 2026-02-11
fix(deepagents): pass on subagent name

christian-bromann:
## Summary

- Pass `name` to `createAgent()` when creating subagents, matching the Python implementation. This causes LangGraph to automatically set `lc_agent_name` in `config.metadata` during tool execution, enabling tools to identify which agent invoked them.
- Add an integration test that verifies `config.metadata.lc_agent_name` is correctly propagated to tools running inside a named subagent.

## Motivation

Users building hierarchical agent systems reported that they couldn't identify which subagent invoked a shared tool, and couldn't pass agent-specific configuration to subagents. The root cause was that the JS `getSubagents()` function wasn't passing the `name` parameter to `createAgent()` — unlike the Python version which does (`create_agent(..., name=agent_["name"])`). Without this, `lc_agent_name` was never set in metadata.


christian-bromann:
Requires fix https://github.com/langchain-ai/langchainjs/pull/9996

PR #226, opened 2026-02-12
Todos management

brettshollenberger:
Stream todos from subagents to parent


christian-bromann (commented):
Thanks for the contribution @brettshollenberger! Can you share a little bit about the motivation for this change? What problems are you trying to solve with this?

PR #234, opened 2026-02-17
fix(deepagents): update summarization behavior

christian-bromann:
This matches the recent changes made in Python.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,7765 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .:
+    devDependencies:
+      '@changesets/changelog-github':
+        specifier: ^0.5.2
+        version: 0.5.2
+      '@changesets/cli':
+        specifier: ^2.29.8
+        version: 2.29.8(@types/node@25.2.3)
+      '@eslint/eslintrc':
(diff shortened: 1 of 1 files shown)


libs/deepagents/src/middleware/summarization.ts
```diff
@@ -687,120 +959,110 @@ ${summary}
...
+            err && typeof err === "object" && "message" in err
+              ? String((err as { message: unknown }).message)
+              : typeof err === "string"
+                ? err
+                : JSON.stringify(err);
+          const isContextOverflow =
+            /prompt.*(too long|too large)|token.*limit|context.*overflow|context.*length|maximum.*token|exceeds.*maximum/i.test(
+              errMsg,
+            );
+
```
hntrl: thoughts on just using ContextOverflowError.isInstance? (i think you would need to rebase the langchain PR if we were doing that

libs/deepagents/src/middleware/summarization.ts
```diff
@@ -234,23 +290,67 @@ export function createSummarizationMiddleware(
...
+   * Called once when the model is first resolved.
+   */
+  function applyModelDefaults(resolvedModel: BaseChatModel): void {
+    if (defaultsComputed) {
+      return;
+    }
+    defaultsComputed = true;
+
+    const defaults = computeSummarizationDefaults(resolvedModel);
+    console.log(
```
hntrl: can we get rid of the logs

PR #236, opened 2026-02-18
feat(deepagents): add local shell backend

christian-bromann:
## Summary

- Adds `LocalShellBackend`, a new backend that extends `FilesystemBackend` with unrestricted local shell command execution via `spawnSync()` — porting the Python `LocalShellBackend` to JS
- Supports configurable timeout, output truncation, custom environment variables, and `inheritEnv` for inheriting the parent process environment
- Includes 18 unit tests covering execution, error handling, timeouts, truncation, env vars, filesystem integration, and virtual mode
- Adds an example at `examples/backends/local-shell-backend.ts`

### DIFF
--- a/internal/standard-tests/src/index.ts
+++ b/internal/standard-tests/src/index.ts
@@ -8,9 +8,6 @@
  * ```
  */
 
-export {
-  sandboxStandardTests,
-  withRetry,
-  type SandboxInstance,
-  type StandardTestsConfig,
-} from "./sandbox.js";
+export { sandboxStandardTests, withRetry } from "./sandbox.js";
+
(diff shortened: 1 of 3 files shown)


libs/deepagents/src/backends/local-shell.ts
```diff
@@ -0,0 +1,242 @@
...
+    if (!command || typeof command !== "string") {
+      return {
+        output: "Error: Command must be a non-empty string.",
+        exitCode: 1,
+        truncated: false,
+      };
+    }
+
+    try {
+      const result = spawnSync(command, {
```
hntrl: do we want this to be async? Think that matches convention for other sandboxes

PR #240, opened 2026-02-20
fix(deepagents): filter invalid content blocks from subagent ToolMessage

### LINKED ISSUE #239: Bug: returnCommandWithStateUpdate fails when lastMessage.content is an array with tool_use blocks
## Description

When `returnCommandWithStateUpdate` constructs a `ToolMessage`, it passes `lastMessage.content` directly. However, when using Anthropic models, `content` can be an array containing `tool_use` blocks. Passing these blocks as `ToolMessage.content` causes errors downstream because `tool_use` blocks are not valid content for a `ToolMessage`.

## Current Behavior

```ts
// src/subagent.ts - returnCommandWithStateUpdate
const lastMessage = messages?.[messages.length - 1];
return new Command({ update: {
    ...stateUpdate,
    messages: [new ToolMessage({
        content: lastMessage?.content || "Task completed",  // ← array with tool_use blocks passed as-is
        tool_call_id: toolCallId,
        name: "task"
    })]
} });
```

When `lastMessage.content` is an array like:
```json
[
  { "type": "text", "text": "Here is the result..." },
  { "type": "tool_use", "id": "...", "name": "...", "input": {...} }
]
```

The `tool_use` blocks get passed into the `ToolMessage`, which is invalid.

## Expected Behavior

`tool_use` blocks should be filtered out from the content array before constructing the `ToolMessage`:

```ts
let content = lastMessage?.content || "Task completed";
if (Array.isArray(content)) {
    content = content.filter((block) => block.type !== "tool_use");
    if (content.length === 0) content = "Task completed";
}
return new Command({ update: {
    ...stateUpdate,
    messages: [new ToolMessage({
        content,
        tool_call_id: toolCallId,
        name: "task"
    })]
} });
```

## Reproduction

1. Create a supervisor with subagents using `createSubAgentMiddleware`
2. Use an Anthropic model (e.g., Claude) that returns array-format content with `tool_use` blocks
3. The subagent's final message content gets passed as-is to `ToolMessage`, causing errors

## Environment

- deepagents: 1.8.0
- @langchain/anthropic: 1.3.3
- @langchain/core: 1.1.25

## Workaround

We are currently using `patch-package` to apply this fix locally.

### LINKED ISSUE #245: Bug: thinking blocks leak into ToolMessage from returnCommandWithStateUpdate
Related to #239 — `returnCommandWithStateUpdate()` passes `lastMessage.content` directly into `ToolMessage.content` without filtering. This also affects `thinking` and `redacted_thinking` blocks (not just `tool_use` blocks covered in #239).

When a subagent uses `ChatAnthropic` with extended thinking enabled, the final `AIMessage.content` contains `thinking` blocks. These get copied into the `ToolMessage`, which the Anthropic API rejects:

```
400 invalid_request_error: messages.2.content.0.tool_result.content.0: Input tag 'thinking' found using 'type' does not match any of the expected tags: 'document', 'image', 'search_result', 'text', 'tool_reference'
```

PR #240 only filters `tool_use` blocks — `thinking` and `redacted_thinking` should also be filtered.

**Versions:** deepagents 1.7.6, @langchain/anthropic 1.3.17

christian-bromann:
## Summary

- Filter out `tool_use`, `thinking`, and `redacted_thinking` blocks from subagent response content in `returnCommandWithStateUpdate` before constructing the `ToolMessage`. When using Anthropic models, `AIMessage.content` can be an array containing these block types, which are invalid as `ToolMessage` content and cause downstream errors.
- Falls back to `"Task completed"` when filtering removes all content blocks.
- Adds four unit tests covering: `tool_use` filtering, `thinking`/`redacted_thinking` filtering, mixed-invalid-block fallback, and string passthrough.

Fixes #239, fixes #245

### DIFF
--- a/libs/deepagents/src/middleware/subagent.test.ts
+++ b/libs/deepagents/src/middleware/subagent.test.ts
@@ -6,7 +6,9 @@ import {
   BaseMessage,
   SystemMessage,
   HumanMessage,
+  ToolMessage,
 } from "@langchain/core/messages";
+import { RunnableLambda } from "@langchain/core/runnables";
 
 import { createDeepAgent } from "../agent.js";
 import { createSkillsMiddleware } from "./skills.js";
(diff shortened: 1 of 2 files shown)


christian-bromann:
> This PR filters `tool_use` blocks but does not filter `thinking` or `redacted_thinking` blocks, which cause the same issue. See #245.

@antonnak thanks for the feedback, added more blocks to the filter list.

PR #242, opened 2026-02-21
feat(deepagents): support direct skill paths as sources in createSkillsMiddleware

alvedder:
## Description

Previously, sources had to be parent directories (e.g. `/skills/`) whose subdirectories each contained a SKILL.md. Passing a direct skill path (e.g. `/skills/my-skill/`) had no effect because `listSkillsFromBackend` scanned for subdirectories inside the source, found none, and returned an empty list.

## Real case

The motivating scenario is when a curated list of skills organised as individual directories under a shared directory. Each skill has a clear responsibility, and different skills are intentionally assigned to the orchestrator agent vs. subagents:

```typescript
// Desired: pass individual skill paths — one per capability
const SHARED_SKILLS = ['/skills/web-research/'];

const ORCHESTRATOR_SKILLS = [
  '/skills/requirement-completeness-validator/',
  '/skills/diff-risk-analyzer/',
];

createDeepAgent({
  skills: [...SHARED_SKILLS, ...ORCHESTRATOR_SKILLS],
  subagents: subagentsConfig.map(agentConfig => ({
    skills: [
      ...SHARED_SKILLS,
      ...agentConfig.skills,
    ]
  }))
});
```

The workaround would be to create nested directories like this:
```
skills/
  web-research/
    web-research/
      SKILL.md
  equirement-completeness-validator/
    equirement-completeness-validator/
      SKILL.md
  ...and so on
```

## What's done

Detection is done by inspecting the lsInfo() result of the source directory: if a SKILL.md file appears directly in the listing, the source itself is treated as the skill directory. Otherwise the existing parent-directory scan runs unchanged.

The two modes can be freely mixed in the sources array:
```ts
sources: [
  '/skills/',               // parent dir: all subdirs with SKILL.md
  '/skills/my-skill/',      // direct path: SKILL.md at its root
 ]
 ```


hntrl (approved):
thanks for the PR! and apologies for the delay in review!

PR #244, opened 2026-02-23
feat(deepagents): add namespace option to StoreBackend

hntrl:
## Summary
This PR simplifies StoreBackend namespacing by removing the extra factory layer and making the namespace an explicit `string[]` option. The backend factory passed to `createDeepAgent` already provides the dynamic context needed, so consumers can compute namespaces there and pass a static array into `StoreBackend` without additional indirection.

Key changes:
- `StoreBackend` now accepts `namespace?: string[]` and validates components for safe characters and non‑empty values.
- The legacy `assistantId` fallback namespace is preserved for backwards compatibility when no custom namespace is provided.
- Removed the earlier namespace factory/context types in favor of the simpler array‑based option.
- Added integration tests that simulate cloud deployment by injecting `store` via invoke config, proving propagation through `createAgent`/`createDeepAgent`.

### DIFF
--- a/.changeset/store-namespace-option.md
+++ b/.changeset/store-namespace-option.md
@@ -1,5 +1,5 @@
 ---
-"deepagents": minor
+"deepagents": patch
 ---
 
 Add `namespace` option to `StoreBackend` for custom store namespace isolation.


christian-bromann (approved):
LGTM 👍

PR #247, opened 2026-02-23
fix(deepagents): CompositeBackend grepRaw/globInfo respect path→backend mapping (#241)

### LINKED ISSUE #241: Composite backend doesn't respect path->backend mapping
Please let me know whether it's expected behavior or not, if the latter then I can submit a PR.

## Description
Composite backend is supposed to route requests to backends based on path mapping, yet `globInfo` and `grepRaw` do not respect the path mapping configuration

## Current Behavior
Both `globInfo` and `grepRaw` follow the same algorithm:
1. try to find a backend by the path mapping configuration and delegate the command
2. if a matching routed backend is not found delegate the command to the default backend
3. **keep delegating to each and every routed backend no matter what the path config says** 

[`grepRaw` implementation](https://github.com/langchain-ai/deepagentsjs/blob/main/libs/deepagents/src/backends/composite.ts#L150-L201)
[`globInfo` implementation](https://github.com/langchain-ai/deepagentsjs/blob/main/libs/deepagents/src/backends/composite.ts#L206-L240)

## Expected Behavior
I'd expect step 3 to be dropped: if no routed backend is found then delegate only to the default backend and return the result. 
Routed backends are configured for a reason and from my perspective they should be used for matching paths only.

## Reproduction
For a setup like this
```ts
new CompositeBackend(workspaceFs, { // a remote sandbox
    'memory/': configsFs, // host machine's filesystem
    'skills/': configsFs, // same
})
```

a tool call like this
```json
"tool_calls": [
  {
    "name": "grep",
    "args": {
      "path": "/workspace",
      "pattern": "index",
      "glob": "**/*"
    },
    "id": "call_eheeJEPNpT70OljJ5eYlrrbe",
    "type": "tool_call"
  }
]
```
will trigger composite backend to search within `configsFs` backend which in my case fails with "EPERM: operation not permitted, scandir '/Library/Bluetooth'" (which is expected on a MacOS machine)

Environment
deepagents: 1.8.0

## Workaround
I extended `CompositeBackend` in my repo and fully overrode `globInfo` and `grepRaw`

christian-bromann:
Fixes #241.

- `grepRaw` and `globInfo` in `CompositeBackend` no longer fan out to every routed backend when the search path doesn't match a route prefix. They now delegate only to the default backend, matching the behavior of all other operations (`read`, `write`, `edit`, `lsInfo`).
- Removed the now-unused `private routes` field (only `sortedRoutes` is needed).
- Updated tests to assert the corrected routing: searching at `/` only queries the default backend; searching with an explicit route prefix (e.g. `/memories/`) correctly queries the routed backend.

## Motivation

When a `CompositeBackend` is configured with path-based routing like:

```ts
new CompositeBackend(workspaceFs, {
    'memory/': configsFs,
    'skills/': configsFs,
})
```

### DIFF
--- a/.changeset/light-cars-report.md
+++ b/.changeset/light-cars-report.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+fix(deepagents): CompositeBackend grepRaw/globInfo respect path→backend mapping (#241)
--- a/libs/deepagents/src/backends/composite.test.ts
+++ b/libs/deepagents/src/backends/composite.test.ts
@@ -155,14 +155,27 @@ describe("CompositeBackend", () => {
       expect(matches1.some((m) => m.path === "/file.txt")).toBe(true);
     }
 
+    // Searching at "/" should NOT search routed backends
     const matches2 = await composite.grepRaw("beta", "/");
     expect(Array.isArray(matches2)).toBe(true);
     if (Array.isArray(matches2)) {
-      expect(matches2.some((m) => m.path === "/memories/readme.md")).toBe(true);
+      expect(matches2.some((m) => m.path === "/memories/readme.md")).toBe(
+        false,
+      );
+    }
+
+    // But searching with the route prefix should find it
+    const matches3 = await composite.grepRaw("beta", "/memories/");
+    expect(Array.isArray(matches3)).toBe(true);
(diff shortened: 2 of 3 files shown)


hntrl:
Closing in favor of #572

PR #258, opened 2026-02-28
feat(fs): add encoding option to filesystem tools

zhisenyang:


christian-bromann (commented):
Changes look good to me. Could we add a unit test that verifies that encoding works properly?

libs/deepagents/src/backends/filesystem.test.ts
```diff
@@ -401,4 +401,61 @@ describe("FilesystemBackend", () => {
...
+    const text = "Hello encoding with utf-16le";
+    const encoding = "utf-16le";
+
+    // 1. Write with utf-16le encoding
+    const writeRes = await backend.write(file, text, encoding);
+    expect(writeRes.error).toBeUndefined();
+
+    // Verify raw file content on disk is actually utf-16le encoded
+    const diskContentRaw = await fs.readFile(file);
+    const decodedDiskContent = Buffer.from(diskContentRaw).toString(encoding);
```
christian-bromann: Let's not use Buffer, instead TextDecoder

libs/deepagents/src/backends/filesystem.test.ts
```diff
@@ -401,4 +401,61 @@ describe("FilesystemBackend", () => {
...
+
+  it("should support encoding parameter for file operations", async () => {
+    const root = tmpDir;
+    const backend = new FilesystemBackend({
+      rootDir: root,
+      virtualMode: false,
+    });
+
+    const file = path.join(root, "encoding-test.txt");
+    const text = "Hello encoding with utf-16le";
```
christian-bromann: Looks like we use a utf-16le encoding but a basic utf-8 text - I think we would get more confidence in the test if we also work with text that requires a different encoding

christian-bromann (commented):
Two more comments

libs/deepagents/src/backends/filesystem.test.ts
zhisenyang: Thanks for the review! Sure, I’ll add a unit test to verify that the encoding works properly.

libs/deepagents/src/backends/filesystem.test.ts
zhisenyang: ok

hntrl (commented):
Closing due to inactivity

PR #261, opened 2026-03-01
feat: add quickjs middleware

hntrl:
## Summary

Adds `@langchain/quickjs` — a sandboxed JavaScript/TypeScript REPL tool for deepagents, powered by QuickJS-NG compiled to WASM via quickjs-emscripten.

## Motivation

LLMs are unreliable at arithmetic, data transformation, and multi-step logic. They hallucinate calculations, lose track of intermediate state, and can't natively express patterns like parallel fan-out or programmatic aggregation. The standard workaround — giving agents a code execution tool — traditionally requires either an unsandboxed Node.js `eval` (a security nightmare) or a heavyweight container/VM.

This PR introduces a third option: a WASM-sandboxed QuickJS interpreter that runs in-process with **no network access, no filesystem access, and no Node.js APIs**. The sandbox boundary is the WebAssembly memory model itself — guest code physically cannot reach host resources unless we explicitly bridge them in.

On top of this sandbox, there are two capabilities that make the REPL useful for deepagents:

### 1. Programmatic Tool Calling (PTC)

Any tool passed to a deepagents agent can be exposed inside the REPL as a typed async function. Instead of the LLM emitting tool calls one at a time through the standard request/response cycle, it **writes code that calls tools directly** — loops, conditionals, error handling, parallel execution, and result transformation all happen in code rather than in the LLM's attention.

```typescript
const urls = ["/users", "/orders", "/products"];
const responses = await Promise.all(
  urls.map(u => tools.httpRequest({ url: "https://api.example.com" + u }))
);
const parsed = responses.map(r => JSON.parse(r));
const summary = { users: parsed[0].length, orders: parsed[1].length, products: parsed[2].length };
await writeFile("/summary.json", JSON.stringify(summary, null, 2));
```

As a specific carve-out of this pattern, PTC enables the **Recursive Language Model (RLM)** architecture: when the `task` tool is exposed inside the REPL, the LLM can spawn sub-agents in parallel via `Promise.all` + `tools.task()`, collect their results, and aggregate programmatically — all within a single `js_eval` call. This turns code into a control plane for multi-agent orchestration.

PTC configuration is progressive: `false` (off), `true` (all tools minus VFS defaults), `string[]` (allowlist), `{ include }` / `{ exclude }` (fine-grained control).

### 2. TypeScript Support & Cross-Eval State Persistence

LLMs naturally generate TypeScript (type annotations, interfaces, `as` casts). Rather than requiring the model to write pure JavaScript, we run an AST-based transform pipeline (acorn + estree-walker + magic-string) that strips TypeScript syntax, hoists top-level declarations to `globalThis` for persistence across evaluations, auto-returns the last expression, and wraps everything in an async IIFE.

This means:
- `const data = await readFile("/data.json")` in eval #1 → `data` is available in eval #2
- Type annotations, interfaces, and generics are silently stripped
- Top-level `await` works naturally
- Parse failures gracefully degrade (the raw code is wrapped and QuickJS reports the syntax error)

### New package: `@langchain/quickjs` (`libs/providers/quickjs/`)

| File | Purpose |
|---|---|
| `types.ts` | `QuickJSMiddlewareOptions`, `ReplSessionOptions`, `ReplResult` |
| `session.ts` | `ReplSession` — lazy WASM runtime lifecycle, VFS bridge, console capture, PTC host functions, buffered writes, timeout management, serialization (`toJSON`/`fromJSON`) |
| `transform.ts` | `transformForEval()` — AST pipeline: strip TS → remove imports/exports → hoist declarations to `globalThis` → auto-return → async IIFE wrap |
| `middleware.ts` | `createQuickJSMiddleware()` — agent integration: injects `js_eval` tool, assembles dynamic system prompt (base + auto-generated PTC type signatures), resolves backend per-thread |
| `utils.ts` | `toCamelCase`, `formatReplResult`, `safeToJsonSchema`, `toolToTypeSignature` |
| `index.ts` | Barrel exports |

**68 unit tests** across 4 test files + **1 integration test** (session: 33, middleware: 9, transform: 16, utils: 10, integration: 1).

Key dependencies: `quickjs-emscripten`, `@jitl/quickjs-ng-wasmfile-release-asyncify` (asyncify variant needed for async host functions), `acorn`, `@sveltejs/acorn-typescript`, `estree-walker`, `magic-string`, `json-schema-to-typescript`, `dedent`.

### `deepagents` core changes

**`libs/deepagents/src/middleware/subagents.ts`** — The `task` tool previously threw if invoked without a `toolCall.id` (which is always absent in PTC context). Now it gracefully returns the sub-agent's last message as a plain string when called via PTC, and returns the normal `Command` with state update when called as a standard tool. This enables the RLM pattern — `tools.task()` inside the REPL returns the sub-agent's response directly to the calling code.

**`libs/deepagents/src/index.ts`** — Exports `StateAndStore` type (needed by the middleware for backend resolution).

### Examples (`examples/repl/`)

- **`data-analysis-agent.ts`** — Agent with QuickJS REPL for data analysis tasks
- **`rlm-agent.ts`** — Recursive Language Model example: PTC-enabled agent that spawns sub-agents inside the REPL via `tools.task()` + `Promise.all`

## Architecture

```
Agent (with middleware)
  │
  ├── wrapModelCall (runs before each LLM call)
  │     ├── Filter PTC tools from agent's toolset
  │     ├── Generate system prompt (base + PTC type signatures)
  │     └── Append to system messages
  │
  └── js_eval tool (injected into agent's tools)
        ├── Get threadId from config.configurable.thread_id
        ├── Resolve backend from state via getCurrentTaskInput()
        ├── ReplSession.getOrCreate(threadId, { backend, tools })
        ├── transformForEval(code) → AST pipeline
        ├── session.eval(transformed, timeout)
        │     ├── ensureStarted() — lazy WASM boot on first call
        │     ├── Set interrupt handler (CPU timeout)
        │     ├── evalCodeAsync in QuickJS WASM
        │     ├── Poll promise state (async timeout)
        │     └── Return { ok, value/error, logs }
        ├── session.flushWrites(backend) — persist buffered writes
        └── formatReplResult → string for LLM

Inside QuickJS WASM sandbox:
  ├── console.{log,warn,error,info,debug} → captured to logs[]
  ├── readFile(path)  → backend.readRaw()
  ├── writeFile(path, content) → buffered to pendingWrites[]
  └── tools.name(input) → tool.invoke() via PTC bridge
```

## Design Decisions

- **WASM variant**: `@jitl/quickjs-ng-wasmfile-release-asyncify` — asyncify is required for `evalCodeAsync` which enables async host functions (VFS reads, PTC tool calls). The `-ng` variant uses QuickJS-NG (actively maintained fork).
- **Session singleton per thread**: Sessions are deduped by `threadId` in a static `Map`. This ensures REPL state persists across multiple `js_eval` calls within an agent thread while preventing cross-thread contamination.
- **Lazy runtime**: The QuickJS WASM runtime is not loaded until the first `.eval()` call. `ReplSession.getOrCreate()` is synchronous — it returns a lightweight handle immediately. This makes sessions safe to create during `wrapModelCall` and survive graph interrupts.
- **Buffered writes**: `writeFile()` inside the sandbox pushes to a `pendingWrites` array synchronously. After eval, the caller flushes writes to the backend. This prevents partial writes from failed scripts and keeps the sandbox deterministic.
- **Session creation in tool handler, not wrapModelCall**: LangGraph's `wrapModelCall` receives a `runtime` object that is intentionally frozen to `{context, writer, interrupt, signal}` — `configurable` (with `thread_id`) is not exposed. Tool handlers receive the full `RunnableConfig` including `config.configurable.thread_id`. Therefore session lookup/creation happens in the `js_eval` tool handler.
- **No env/secrets system**: The original design included environment variable isolation with secret management (plain strings, tool-restricted values, opaque secret placeholders). This was removed to reduce complexity — secrets management is better handled at the application/deployment layer rather than inside the REPL sandbox.
- **No heap persistence**: quickjs-emscripten doesn't expose a way to create a runtime from existing WASM memory, so sessions are ephemeral to the process lifetime (not serializable to LangGraph checkpoints). The `toJSON()`/`fromJSON()` methods serialize only the session id — on restore, `fromJSON()` reconnects to the live runtime if it still exists, or creates a fresh session.
- **Negative timeout = no timeout**: Setting `executionTimeoutMs` to a negative value disables both the CPU interrupt handler and the promise polling deadline, allowing unbounded execution for long-running PTC workflows.
- **Graceful transform fallback**: If the AST parser fails (malformed code), the raw code is wrapped in an async IIFE anyway. QuickJS then reports the actual syntax error, which is more useful than an opaque parse failure.

### DIFF
--- a/libs/providers/quickjs/README.md
+++ b/libs/providers/quickjs/README.md
@@ -23,7 +23,9 @@ const agent = createDeepAgent({
 });
 
 const result = await agent.invoke({
-  messages: [{ role: "user", content: "Calculate the first 20 Fibonacci numbers" }],
+  messages: [
+    { role: "user", content: "Calculate the first 20 Fibonacci numbers" },
+  ],
 });
 ```
 
@@ -32,7 +34,7 @@ The agent now has a `js_eval` tool. It can write and execute JavaScript/TypeScri
 ```typescript
 // Call 1: the agent writes
 var fibs = [0, 1];
-for (let i = 2; i < 20; i++) fibs.push(fibs[i-1] + fibs[i-2]);
+for (let i = 2; i < 20; i++) fibs.push(fibs[i - 1] + fibs[i - 2]);
 console.log(fibs);
 
 // Call 2: state persists — `fibs` is still available
@@ -80,21 +82,21 @@ Inside the REPL, the agent can then write:
 ```typescript
 const urls = ["/users", "/orders", "/products"];
 const results = await Promise.all(
-  urls.map(u => tools.httpRequest({ url: "https://api.example.com" + u }))
+  urls.map((u) => tools.httpRequest({ url: "https://api.example.com" + u })),
 );
-const parsed = results.map(r => JSON.parse(r));
+const parsed = results.map((r) => JSON.parse(r));
 console.log(`Users: ${parsed[0].length}, Orders: ${parsed[1].length}`);
 ```
 
 PTC configuration is progressive:
 
-| Value | Behavior |
-|---|---|
-| `false` | Disabled (default) |
-| `true` | All agent tools except VFS builtins |
-| `string[]` | Only these tools |
-| `{ include: string[] }` | Only these tools |
-| `{ exclude: string[] }` | All tools except these |
+| Value                   | Behavior                            |
+| ----------------------- | ----------------------------------- |
+| `false`                 | Disabled (default)                  |
+| `true`                  | All agent tools except VFS builtins |
+| `string[]`              | Only these tools                    |
+| `{ include: string[] }` | Only these tools                    |
+| `{ exclude: string[] }` | All tools except these              |
 
 ### Recursive Language Model (RLM)
 
@@ -103,7 +105,13 @@ When the `task` tool is exposed via PTC, the agent can spawn sub-agents in paral
 ```typescript
 const agent = createDeepAgent({
   model: "claude-sonnet-4-5-20250929",
-  subagents: [{ name: "general-purpose", description: "Research agent", systemPrompt: "..." }],
+  subagents: [
+    {
+      name: "general-purpose",
+      description: "Research agent",
+      systemPrompt: "...",
+    },
+  ],
   middleware: [createQuickJSMiddleware({ ptc: ["task"] })],
 });
 ```
@@ -113,12 +121,12 @@ The agent then writes code like:
 ```typescript
 const topics = ["quantum computing", "fusion energy", "CRISPR"];
 const results = await Promise.all(
-  topics.map(topic =>
+  topics.map((topic) =>
     tools.task({
       description: `Research ${topic} in depth`,
       subagentType: "general-purpose",
-    })
-  )
+    }),
+  ),
 );
 const report = topics.map((t, i) => `## ${t}\n${results[i]}`).join("\n\n");
 await writeFile("/research.md", report);
@@ -134,10 +142,10 @@ Creates a middleware that adds the `js_eval` tool to your agent.
 interface QuickJSMiddlewareOptions {
   backend?: BackendProtocol | BackendFactory; // File I/O backend (default: StateBackend)
   ptc?: boolean | string[] | { include: string[] } | { exclude: string[] }; // PTC config
-  memoryLimitBytes?: number;     // Default: 50MB
-  maxStackSizeBytes?: number;    // Default: 320KB
-  executionTimeoutMs?: number;   // Default: 30s (-1 to disable)
-  systemPrompt?: string | null;  // Override the built-in REPL system prompt
+  memoryLimitBytes?: number; // Default: 50MB
+  maxStackSizeBytes?: number; // Default: 320KB
+  executionTimeoutMs?: number; // Default: 30s (-1 to disable)
+  systemPrompt?: string | null; // Override the built-in REPL system prompt
 }
 ```
 


examples/repl/data-analysis-agent.ts
```diff
@@ -0,0 +1,51 @@
...
+import { createQuickJSMiddleware } from "@langchain/quickjs";
+
+const model = new ChatAnthropic({
+  model: "claude-sonnet-4-20250514",
+  temperature: 0,
+});
+
+const agent = createDeepAgent({
+  model,
+  systemPrompt: dedent`
```
christian-bromann, line 28: Didn't we add some primitives to the `langchain` package?

examples/repl/secret-env-agent.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Configuration:
+ * - `env.API_KEY` → secret, only allowed in `http_request` tool
+ * - `env.BASE_URL` → plain, restricted to `http_request` tool
+ * - `env.APP_NAME` → plain, unrestricted (available everywhere)
+ */
+import "dotenv/config";
+import dedent from "dedent";
+import { z } from "zod";
+import { tool } from "langchain";
+import { HumanMessage } from "@langchain/core/messages";
```
christian-bromann:
```suggestion
import { tool, HumanMessage } from "langchain";
```

libs/providers/quickjs/src/middleware.ts
```diff
@@ -0,0 +1,308 @@
...
+import { z } from "zod/v4";
+import type { StructuredToolInterface } from "@langchain/core/tools";
+import {
+  StateBackend,
+  type BackendProtocol,
+  type BackendFactory,
+  type StateAndStore,
+} from "deepagents";
+
+import dedent from "dedent";
```
christian-bromann, line 25: Same here, shouldn't we use https://reference.langchain.com/javascript/langchain/index/context ?

libs/providers/quickjs/src/middleware.ts
```diff
@@ -0,0 +1,308 @@
...
+          ),
+      }),
+    },
+  );
+
+  return createMiddleware({
+    name: "QuickJSMiddleware",
+    tools: [jsEvalTool],
+    wrapModelCall: async (request, handler) => {
+      const threadId = request.runtime.configurable?.thread_id || "__default__";
```
christian-bromann: Can we make `__default__` a constant with a comment on what it is used for since we also use it in line 248.

libs/providers/quickjs/src/session.ts
```diff
@@ -0,0 +1,465 @@
...
+ * then `session.eval(code, timeoutMs)` to execute code.
+ */
+export class ReplSession {
+  private static sessions = new Map<string, ReplSession>();
+
+  readonly runtime: QuickJSAsyncRuntime;
+  readonly context: QuickJSAsyncContext;
+  readonly logs: string[] = [];
+
+  private _backend: BackendProtocol | null = null;
```
christian-bromann, line 85: Since we have a getter for this
```suggestion
  private #backend: BackendProtocol | null = null;
```

christian-bromann: I guess since this is an internal class we don't expose to users it doesn't really matter, so feel free to ignore

libs/providers/quickjs/src/session.ts
```diff
@@ -0,0 +1,465 @@
...
+export const DEFAULT_MAX_STACK_SIZE = 320 * 1024;
+export const DEFAULT_EXECUTION_TIMEOUT = 30_000;
+
+let asyncModulePromise: Promise<any> | undefined;
+
+async function getAsyncModule() {
+  if (!asyncModulePromise) {
+    asyncModulePromise = (async () => {
+      const variant =
+        await import("@jitl/quickjs-ng-wasmfile-release-asyncify");
```
christian-bromann, line 49: Is there a specific reason we dynamically import here instead of statically at the top?

christian-bromann (commented):
Some minor nits. Awesome stuff 🙌

libs/providers/quickjs/src/session.ts
hntrl: Has something to do with how web assembly gets loaded inside of that module. This is the pattern they recommend

PR #273, opened 2026-03-04
feat(subagents): Add structured response support for subagents

maahir30:
Adds the ability for subagents to return structured (schema-validated) responses, both statically at definition time and dynamically at invocation time.

### What this enables

**Static `responseFormat`** — When defining a subagent spec, you can now set a `responseFormat` (e.g. `toolStrategy(zodSchema)`) that gets baked into the subagent's agent graph at build time. Every invocation of that subagent will enforce the schema.

**Dynamic `response_schema`** — The supervisor can pass a `response_schema` (a plain JSON Schema object) in its `task` tool call. When this happens, the middleware builds a fresh agent graph on the fly with that schema as the `responseFormat`. This gives the supervisor per-invocation control over the subagent's output shape without needing to predefine every possible format.

### How it works

- When a subagent produces a `structuredResponse`, `returnCommandWithStateUpdate` now JSON-serializes it into the `ToolMessage` content sent back to the supervisor, instead of extracting the last message text. This gives the supervisor predictable, parseable data.
- Dynamic schema creation only applies to spec-based subagents that don't already have a static `responseFormat`. For compiled subagents or those with a static format, the `response_schema` parameter is ignored and the pre-built graph is used as-is.

### Changes

- **`subagents.ts`**: Added `responseFormat` to the `SubAgent` interface, `response_schema` to the `task` tool schema, on-the-fly agent creation logic, and `structuredResponse` prioritization in `returnCommandWithStateUpdate`
- **`subagent.test.ts`**: Unit tests for structured response serialization, fallback behavior, and dynamic schema bypass paths
- **`subagents.int.test.ts`**: Integration tests for both dynamic `response_schema` and static `responseFormat` with a real LLM

### DIFF
--- a/libs/deepagents/src/middleware/subagents.ts
+++ b/libs/deepagents/src/middleware/subagents.ts
@@ -74,6 +74,7 @@ When using the Task tool, you must specify a subagent_type parameter to select w
 5. Clearly tell the agent whether you expect it to create content, perform analysis, or just do research (search, file reads, web fetches, etc.), since it is not aware of the user's intent
 6. If the agent description mentions that it should be used proactively, then you should try your best to use it without the user having to ask for it first. Use your judgement.
 7. When only the general-purpose agent is provided, you should use it for all tasks. It is great for isolating context and token usage, and completing specific, complex tasks, as it has all the same capabilities as the main agent.
+8. Use the \`response_schema\` parameter when you need structured JSON from subagents, especially for parallel comparison or data aggregation. When you receive JSON from any subagent, incorporate the data naturally into your response.
 
 ### Example usage of the general-purpose agent:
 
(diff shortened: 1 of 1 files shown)


libs/deepagents/src/middleware/subagents.ts
```diff
@@ -601,6 +706,16 @@ function createTaskTool(options: {
...
+        response_schema: z
+          .object({
+            type: z.string(),
+            properties: z.record(z.string(), z.unknown()).optional(),
+            required: z.array(z.string()).optional(),
+          })
+          .optional()
+          .describe(
+            "Optional JSON Schema to enforce structured output from the subagent. Only use when you need the subagent to return data in a specific structure.",
+          ),
```
hntrl, line 718: weren't we going to nix this behavior just to make responseFormat configurable? E.g. just the static case not the dynamic case for now

Issue #277, opened 2026-03-05
Large tool results not checked for content length if using non-string content blocks


bracesproul:
the logic which checks the size of the tool result only considers `content` blocks which are strings, not multi-type content blocks

this means if you have a complex content block with a `text` field that contains 1 billion characters, it'll get sent right back to the agent

i think the fix here is to join all content blocks with type `text`, and check the lengths of those

`libs/deepagents/src/middleware/fs.ts` `L857`

PR #278, opened 2026-03-05
fix: handle multi-type content blocks in tool result size checking

bracesproul:
## Description
Fixes a bug where the tool result eviction logic only checked `content` when it was a plain string, ignoring array-based content blocks (e.g., `[{ type: "text", text: "..." }]`). This meant that a tool result with complex content blocks containing arbitrarily large `text` fields would bypass the size limit and be sent directly back to the agent.

The fix introduces an `extractTextContent` helper that handles both content formats:
- **String content**: returned as-is (existing behavior)
- **Array content blocks**: joins all blocks with `type === "text"` and checks the combined length

Changes:
- Added `extractTextContent()` utility function in `fs.ts`
- Updated `processToolMessage()` to use `extractTextContent()` instead of only checking string content
- Added 6 unit tests for `extractTextContent`
- Exported `extractTextContent` from the middleware index

Resolves https://github.com/langchain-ai/deepagentsjs/issues/277

## Test Plan
- [x] All 499 existing + new tests pass (`pnpm test`)
- [x] TypeScript compilation passes (`pnpm exec tsc --noEmit`)
- [x] Lint passes (`pnpm lint`)
- [x] Format passes (`pnpm format`)
- [ ] Verify that tool results with array content blocks containing large text are properly evicted
- [ ] Verify that tool results with string content still work as before

### DIFF
--- a/libs/deepagents/src/middleware/fs.ts
+++ b/libs/deepagents/src/middleware/fs.ts
@@ -136,6 +136,34 @@ export function createContentPreview(
 }
 
 /**
+ * Extract joined text from message content (string or array of content blocks).
+ */
+export function extractTextContent(
+  content: string | Array<Record<string, unknown>>,
+): string | null {
+  if (typeof content === "string") {
+    return content;
+  }
+  if (Array.isArray(content)) {
+    const textParts: string[] = [];
+    for (const block of content) {
+      if (
+        typeof block === "object" &&
+        block !== null &&
+        block.type === "text" &&
+        typeof block.text === "string"
+      ) {
+        textParts.push(block.text);
+      }
+    }
+    if (textParts.length > 0) {
+      return textParts.join("");
+    }
+  }
+  return null;
+}
+
+/**
  * required for type inference
  */
 import type * as _zodTypes from "@langchain/core/utils/types";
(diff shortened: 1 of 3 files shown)


libs/deepagents/src/middleware/fs.ts
```diff
@@ -853,9 +881,10 @@ export function createFilesystemMiddleware(
         msg: ToolMessage,
         toolTokenLimitBeforeEvict: number,
       ) {
+        const textContent = extractTextContent(msg.content);
```
christian-bromann, line 884: Shouldn't this suffice?
```suggestion
        const textContent = msg.text;
```

libs/deepagents/src/middleware/fs.ts
```diff
@@ -135,6 +135,34 @@ export function createContentPreview(
   return headSample + truncationNotice + tailSample;
 }
 
+/**
+ * Extract joined text from message content (string or array of content blocks).
+ */
+export function extractTextContent(
```
christian-bromann, line 141: Which would allow use to remove this function

christian-bromann (changes requested):

Issue #82, opened 2025-12-16
Agent can enter an infinite loop when handling large tool results with long lines


KDKHD:
### Summary

When a tool returns a large result composed of a small number of very long lines, the agent can enter an infinite eviction/read loop.

### Root Cause

The issue arises from an interaction between:

- `toolTokenLimitBeforeEvict` (default: `20000`)
- Eviction of large tool results to the filesystem
- The `read_file` tool’s **line-based** limit (default: `2000` lines)

### Detailed Explanation

1. By default, `toolTokenLimitBeforeEvict` is set to `20000`.

2. If a tool returns a result containing **very long lines** (e.g. 5 lines totaling ~100,000 characters), the result exceeds the eviction threshold  
   (`100,000 chars > 20,000 tokens * 4 chars/token`) and is evicted to the filesystem.  
   See:  
   https://github.com/langchain-ai/deepagentsjs/blob/main/src/middleware/fs.ts#L481

3. The agent then attempts to read the evicted content using the `read_file` tool.

4. `read_file` applies a **line limit** (default: 2000 lines):  
   https://github.com/langchain-ai/deepagentsjs/blob/main/src/middleware/fs.ts#L193

5. Because the file contains only a few lines (e.g. 5), all ~100,000 characters are returned in a single `read_file` result.

6. The `read_file` tool result now again exceeds the eviction threshold and is evicted to the filesystem.

7. The agent repeats the process, resulting in an infinite loop:

### Trace

Reproduction trace:  
https://smith.langchain.com/public/5ccfa3e5-07f0-4572-b205-22ae3d3d5689/r

### Possible Solutions

- Change `read_file` to enforce a **character limit** instead of (or in addition to) a line limit
- Exempt `read_file` results from eviction. It doesn't make sense for the read_file tool to result in evictions, as this would result in files being duplicated.
- Alternatively, cap the returned content from `read_file` based on token or character count to guarantee forward progress

mrblakessinger-rgb:
The main issue here is that file eviction without progress tracking creates a closed feedback loop. When the read tool pulls from an evicted file and re-triggers the token threshold, the agent loses track of context progress and just spins on the same offset. Adding a state delta check on read operations (to verify unique tokens are actually being ingested before re-evaluating eviction) breaks that loop early.

[...]

PR #281, opened 2026-03-06
fix: add truncation to grep/glob/ls tool results

### LINKED ISSUE #272: grep/glob/ls tool results can exceed context window despite eviction
## Problem

`grep`, `glob`, and `ls` can return hundreds of KB (observed: 772KB from a single grep). The eviction mechanism saves large results to the filesystem, but doesn't cap tool output — so the oversized result still enters the message history and causes model retries or blank responses before eviction can help.

`read_file` already has character-based truncation (#172). The same treatment is missing for search tools.

## Proposed solution

Add an optional character/token cap to `grep`, `glob`, and `ls` tool results — matching the `TOOL_RESULT_TOKEN_LIMIT * 4` threshold already used by eviction. When results exceed the cap, truncate and append guidance like:

```
... [results truncated — try narrowing your search with more specific parameters]
```

This is consistent with `read_file`'s existing truncation behavior (#172).

## Related

- #82 — infinite loop from large `read_file` results (fixed by #172, same pattern)
- #172 — added char truncation to `read_file` (merged, but grep/glob/ls not covered)
- #228 — attempted post-summarization emergency truncation for 668KB glob (closed in favor of #234)
- #234 — updated summarization behavior (merged, but no tool-level caps)
- #248 — preserved ToolMessage metadata during eviction (merged, related infra)
- #250 — OOM fix in conversation history offloading (merged, related)

## Workaround

Custom middleware that intercepts `grep`/`glob`/`ls` `ToolMessage` results and truncates at 80K chars using `createMiddleware({ wrapToolCall })`. Works but shouldn't be necessary — the built-in tools should respect the same limits as `read_file`.

JadenKim-dev:
## Summary

Fixes #272

`grep`, `glob`, and `ls` tools can return hundreds of KB of results (observed: 772KB from a single grep), which can exceed context limits and cause model retries or blank responses before eviction can help.

This PR adds truncation at the `TOOL_RESULT_TOKEN_LIMIT` (20K tokens ~80KB) threshold, consistent with `read_file`'s existing behavior from #172 and **matching the Python version's implementation**.

## Changes

- Import `truncateIfTooLong` utility in `fs.ts` middleware
- Apply truncation to `ls`, `glob`, and `grep` tool results
- Truncate arrays before joining to preserve complete lines
- Add guidance message when results are truncated: `... [results truncated, try being more specific with your parameters]`

**Architecture**: The truncation happens in the middleware layer (matching Python's architecture), not in the backend, so backends return full results and middleware applies the limit. This follows the same pattern as the Python implementation in `deepagents/middleware/filesystem.py`.

## Test plan

- [x] Existing tests pass (`utils.test.ts` already covers `truncateIfTooLong` utility at lines 290-314)
- [x] Implementation follows established pattern from `read_file` truncation (#172)
- [x] Matches Python version's architecture and behavior

## Related

- #82 — infinite loop from large tool results (fixed by #172 for `read_file`)
- #172 — added char truncation to `read_file` (merged, same pattern used here)
- #228 — context overflow from 668KB glob (attempted fix via post-summarization truncation)

### DIFF
--- a/.changeset/loud-pigs-clap.md
+++ b/.changeset/loud-pigs-clap.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+fix: add truncation to grep/glob/ls tool results


libs/deepagents/src/middleware/fs.test.ts
```diff
@@ -805,5 +805,10 @@ describe("createFilesystemMiddleware", () => {
       expect(parsed.file_path).toBe("/app/test.c");
       expect(parsed.content).toBe("");
     });
+
+    // Note: truncateIfTooLong is tested in backends/utils.test.ts (lines 290-314)
+    // The truncation logic for ls, grep, and glob tools uses this utility function
+    // and is covered by those existing tests. Full integration tests with LangGraph
+    // context would require complex setup and are better suited for end-to-end testing.
```
christian-bromann: I don't think we need this comment
```suggestion
```

christian-bromann (commented):
Can we add a test for grep/glob/ls results being too long?

libs/deepagents/src/middleware/fs.test.ts
JadenKim-dev: oh I was working on the test code, and it seems the agent committed it arbitrarily. I'll add the rest and push it.

PR #289, opened 2026-03-09
fix(deepagents): move uuid from devDependencies to dependencies

alvedder:
`uuid` package is actually used in runtime by [summarization.ts](https://github.com/langchain-ai/deepagentsjs/blob/main/libs/deepagents/src/middleware/summarization.ts#L44) but specified as a dev dependency.

This breaks packages installed via pnpm that have `deepagents` in dependencies:

```log
import { v4 } from "uuid";
         ^^
SyntaxError: The requested module 'uuid' does not provide an export named 'v4'
    at #asyncInstantiate (node:internal/modules/esm/module_job:319:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:422:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:654:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

Note: if installed via npm it works fine, probably because `npm` keeps the packages flat and resolves `uuid` to the version used by any other dependant package.

Solution: move `uuid` to `dependencies` map.

### DIFF
--- a/.changeset/long-tables-press.md
+++ b/.changeset/long-tables-press.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+fix(deepagents): move uuid from devDependencies to dependencies


christian-bromann (approved):
👍 

Thank you!

PR #291, opened 2026-03-09
feat(acp): add authMethods to InitializeResponse for ACP registry compliance

christian-bromann:
## Summary

- Add `authMethods` to the ACP `InitializeResponse`, fixing the registry CI failure ("No authMethods in response")
- Introduce typed interfaces (`ACPAuthMethod`, `ACPAuthMethodAgent`, `ACPAuthMethodEnvVar`, `ACPAuthMethodTerminal`) matching the [ACP auth methods spec](https://agentclientprotocol.com/rfds/auth-methods)
- Make `authMethods` configurable via `DeepAgentsServerOptions` so users building custom `DeepAgentsServer` instances can define their own auth flows

Defaults to Anthropic/OpenAI `env_var` methods plus a generic agent-type fallback when no `authMethods` are provided:

```typescript
const server = new DeepAgentsServer({
  agents: { name: "my-agent" },
  authMethods: [
    {
      id: "my-login",
      name: "Login",
      type: "terminal",
      args: ["--login"],
    },
    {
      id: "api-key",
      name: "API Key",
      type: "env_var",
      vars: [{ name: "MY_API_KEY" }],
      link: "https://my-service.com/keys",
    },
  ],
});
```

### DIFF
--- a/libs/acp/src/types.ts
+++ b/libs/acp/src/types.ts
@@ -71,7 +71,6 @@ export interface ACPAuthMethodEnvVar {
   id: string;
   name: string;
   type: "env_var";
-  description?: string;
   /** URL where the user can obtain credentials */
   link?: string;
   /** Environment variables required for this auth method */
@@ -86,7 +85,6 @@ export interface ACPAuthMethodTerminal {
   id: string;
   name: string;
   type: "terminal";
-  description?: string;
   /** Additional arguments passed when running the agent binary for login */
   args?: string[];
   /** Additional environment variables set during the login invocation */


libs/acp/src/types.ts
```diff
@@ -36,6 +36,92 @@ export interface DeepAgentConfig extends CreateDeepAgentParams {
...
+
+/**
+ * Env-var auth method — credentials are passed as environment variables.
+ * Clients can show input UI for each variable and restart the agent with them set.
+ */
+export interface ACPAuthMethodEnvVar {
+  id: string;
+  name: string;
+  type: "env_var";
+  description?: string;
```
colifran: Just double checking - I don't see an optional description called out in the linked documentation. Is this okay here?

libs/acp/src/types.ts
```diff
@@ -36,6 +36,92 @@ export interface DeepAgentConfig extends CreateDeepAgentParams {
...
+
+/**
+ * Terminal auth method — requires an interactive terminal for login (e.g. TUI).
+ * Only used when the client advertises `auth.terminal` capability.
+ */
+export interface ACPAuthMethodTerminal {
+  id: string;
+  name: string;
+  type: "terminal";
+  description?: string;
```
colifran: Just double checking here too - I see args and env called out as optional but not description. Should this be required?

PR #293, opened 2026-03-10
fix: update various issues in QuickJS REPL

maahir30:
## Summary

Fixes three bugs in the QuickJS REPL transformer discovered via structured-output eval traces. In the eval runs, these caused 4 unnecessary retries (~5K wasted tokens) from semicolon errors, and ~600 tokens of data re-embedding per cell.

### Semicolons in auto-return cause SyntaxError

- The transformer wraps the last expression in `return (...)` to capture its value
- It was using the `ExpressionStatement` node range, which includes the trailing semicolon
- So `expr;` became `return (expr;)` — a syntax error inside the parens
- The LLM had to waste 4 retries discovering that removing all semicolons was the workaround
- Fixed by using two different AST positions for the wrapping:
  - `ExpressionStatement.start` for the opening `return (` — preserves any grouping parens like `({...})`
  - The inner `expression.end` for the closing `)` — sits before the semicolon in the AST
- Now `expr;` becomes `return (expr);` — the semicolon stays outside the parens

### TypeScript not stripped from variable initializers

- The transformer strips TS in two phases: Phase 1 hoists variable declarations to `globalThis`, Phase 2 walks all other statements and strips TS annotations
- Phase 2 explicitly skips `VariableDeclaration` nodes, assuming Phase 1 handles them
- Phase 1 delegates to `extractCleanInit`, which was supposed to walk the init AST subtree and strip TS — but it only sliced the raw source text and returned it untouched
- So `const data = JSON.parse(raw) as { n: number }` would emit the `as { n: number }` into QuickJS verbatim
- Fixed by adding `extractCleanSource`:
  - Slices the init source into a separate MagicString copy
  - Walks the AST subtree over it, stripping all TS annotations
  - Adjusts every removal by an offset (`node.start`) to convert between absolute AST positions and the sliced MagicString's zero-based positions
- The same helper is also used for destructuring patterns, which had the same gap

### LLM re-embeds data instead of referencing REPL state

- After a `Promise.all` cell that stored results in a variable like `parsed`, the LLM was copying the entire JSON response (~600 tokens) into the next `js_eval` call instead of just writing `parsed.map(...)`
- Variables persist across cells via `globalThis`, but nothing reminded the LLM of that at decision time
- Fixed with two reinforcements:
  - A system prompt hard rule: "reference variables by name instead of re-embedding data as inline JSON literals"
  - A state hint appended to every tool response listing which variables are available in the next cell (`↳ State: results, parsed, cities (available in next cell)`)
- The transformer now collects declared names during hoisting, the session accumulates them across evals, and `formatReplResult` includes them in the output

### DIFF
--- a/libs/providers/quickjs/src/middleware.int.test.ts
+++ b/libs/providers/quickjs/src/middleware.int.test.ts
@@ -83,8 +83,8 @@ describe("QuickJS REPL integration", () => {
           messages: [
             new HumanMessage(
               "Using js_eval, first store this data: " +
-              '[{name: "a", value: 10}, {name: "b", value: 20}, {name: "c", value: 30}]. ' +
-              "Then in a separate js_eval call, compute the sum of all the values.",
+                '[{name: "a", value: 10}, {name: "b", value: 20}, {name: "c", value: 30}]. ' +
+                "Then in a separate js_eval call, compute the sum of all the values.",
             ),
           ],
         },
@@ -96,8 +96,8 @@ describe("QuickJS REPL integration", () => {
 
       // Verify the second js_eval call does not re-embed the array literal
       const aiMessages = result.messages.filter(AIMessage.isInstance);
-      const jsEvalCalls = aiMessages.flatMap(
-        (msg) => (msg.tool_calls || []).filter((tc) => tc.name === "js_eval"),
+      const jsEvalCalls = aiMessages.flatMap((msg) =>
+        (msg.tool_calls || []).filter((tc) => tc.name === "js_eval"),
       );
       expect(jsEvalCalls.length).toBeGreaterThanOrEqual(2);
 
@@ -106,7 +106,8 @@ describe("QuickJS REPL integration", () => {
       expect(secondCallCode).not.toContain('"name": "b"');
 
       // Verify the computation succeeded (10 + 20 + 30 = 60)
-      const lastToolContent = toolMessages[toolMessages.length - 1].content as string;
+      const lastToolContent = toolMessages[toolMessages.length - 1]
+        .content as string;
       expect(lastToolContent).toContain("60");
     },
   );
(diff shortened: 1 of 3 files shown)


libs/providers/quickjs/src/middleware.int.test.ts
```diff
@@ -56,4 +56,61 @@ describe("QuickJS REPL integration", () => {
       expect(session).not.toBeNull();
     },
   );
+
+  it(
+    "should reference variables from prior cells instead of re-embedding data",
```
hntrl, line 61: doesn't this test already exist?

libs/providers/quickjs/src/middleware.ts
```diff
@@ -76,6 +76,7 @@ const REPL_SYSTEM_PROMPT = dedent`
   - **No network, no filesystem** — only the helpers below. Do not attempt \`fetch\`, \`require\`, or \`import\`.
   - **Cite your sources** — when reporting values from files, include the path and key/index so the user can verify.
   - **Use console.log()** for output — it is captured and returned. \`console.warn()\` and \`console.error()\` are also available.
+  - **Reuse state from previous cells** — variables, functions, and results from earlier \`js_eval\` calls persist via \`globalThis\`. Reference them by name in follow-up cells instead of re-embedding data as inline JSON literals. The tool response lists available variable names after each cell.
```
hntrl: we don't want the agent to start writing closures with globalThis -- we extrapolate to globalThis since its a quirk of this implementation

libs/providers/quickjs/src/middleware.int.test.ts
maahir30: the difference between this test and the one above is that this test checks that the second tool call's code referenced the variable instead of re-embedding it. the issue that the evals had. i was gonna add it to the test above but `var x = 99` is too simple to trigger re-embedding (the LLM would never copy a single number).

libs/providers/quickjs/src/middleware.ts
maahir30: oops sorry did not realize that you could just access them normally. I'll fix

PR #298, opened 2026-03-11
feat(deepagents): support multimodal files for backends

colifran:
### Description

Adds support for binary and multimodal files (images, PDFs, audio, video, etc.) across all backend implementations, with a versioned protocol layer that returns structured results instead of plain values or arrays.

#### Protocol
  - Introduced BackendProtocolV2 with structured Result return types:
    - ReadResult — for read() operations
    - ReadRawResult — for readRaw() operations
    - GrepResult — for grepRaw() operations
    - LsResult — for lsInfo() operations
    - GlobResult — for globInfo() operations
  - Introduced SandboxBackendProtocolV2 extending BackendProtocolV2; deprecated SandboxBackendProtocol
  - Deprecated v1 interfaces (BackendProtocol, SandboxBackendProtocol) moved to v1/; current interfaces in v2/ for clear separation
  - Added AnyBackendProtocol union type and adaptBackendProtocol() for backward compatibility — public APIs accept either version and adapt internally
  - FileData split into FileDataV1 (legacy line array) and FileDataV2 (single string + mimeType, supports base64)
  - All Result types follow consistent pattern: { error?: string, [data]?: T } for explicit error propagation

#### Binary File Support
  - All backends (State, Store, Filesystem, Sandbox, Composite) store and retrieve binary files as base64
  - read_file tool returns typed multimodal content blocks (image, audio, video, file) for binary files
  - Binary reads capped at 10MB to stay within provider inline limits
  - MIME type detection via file extension, stored with v2 FileData

#### Middleware
  - Updated skills.ts middleware to handle LsResult from backend operations
  - Updated fs.ts middleware to handle LsResult and GlobResult with proper error checking

#### Provider Updates

  - QuickJS REPL: public API accepts AnyBackendProtocol, adapts to v2 internally via adaptBackendProtocol() — fully backward compatible
  - Node VFS: VfsSandbox updated to return LsResult and GlobResult instead of bare arrays

#### Backward Compatibility
  - v1 backends continue to work everywhere via adaptBackendProtocol() runtime detection
  - v2 backends correctly read and operate on existing v1 FileData (string[] content)
  - No breaking changes to public APIs — all accept AnyBackendProtocol

### Tests
  - All backend tests updated to assert on Result types: composite.test.ts, filesystem.test.ts, state.test.ts, store.test.ts, sandbox.test.ts, local-shell.test.ts
  - Added readRaw() tests returning ReadRawResult across State, LocalShell, Composite, and Node VFS backends
  - utils.test.ts updated to test adaptBackendProtocol() with all Result types, including v1→v2 wrapping
  - Middleware tests (fs.test.ts, skills.test.ts) updated to mock backends returning Result types
  - New binary tests for State and Store backends covering upload, download, round-trip, read, grep, and pagination
  - New multimodal content block tests for read_file (image, audio, video, PDF, size limit enforcement)
  - Sandbox test: readRaw() now returns error in Result object instead of throwing
  - Added v1→v2 backward compatibility tests for State and Store backends (mixed v1/v2 data)
  - Manual e2e testing scripts for all 5 backend configurations plus v1→v2 migration simulation

### Example:
multimodal

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -105,8 +105,8 @@ importers:
         specifier: workspace:*
         version: link:../../libs/deepagents
       langchain:
-        specifier: ^1.2.26
-        version: 1.2.26(@langchain/core@1.1.32(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0)
+        specifier: ^1.2.34
+        version: 1.2.34(@langchain/core@1.1.33(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0)
       langsmith:
         specifier: ^0.5.4
         version: 0.5.7(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))
@@ -159,8 +159,8 @@ importers:
         specifier: workspace:*
         version: link:../../libs/deepagents
       langchain:
-        specifier: ^1.2.26
-        version: 1.2.26(@langchain/core@1.1.32(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0)
+        specifier: ^1.2.34
+        version: 1.2.34(@langchain/core@1.1.33(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0)
       langsmith:
         specifier: ^0.5.4
         version: 0.5.7(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))
@@ -180,8 +180,8 @@ importers:
         specifier: workspace:*
         version: link:../../libs/deepagents
       langchain:
-        specifier: ^1.2.26
-        version: 1.2.26(@langchain/core@1.1.32(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0)
+        specifier: ^1.2.34
+        version: 1.2.34(@langchain/core@1.1.33(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0)
       langsmith:
         specifier: ^0.5.4
         version: 0.5.7(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))
@@ -360,8 +360,8 @@ importers:
         specifier: ^3.3.3
         version: 3.3.3
       langchain:
-        specifier: 1.2.26
-        version: 1.2.26(@langchain/core@1.1.26(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0)
+        specifier: 1.2.34
+        version: 1.2.34(@langchain/core@1.1.33(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0)
       micromatch:
         specifier: ^4.0.8
         version: 4.0.8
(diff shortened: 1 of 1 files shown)


libs/deepagents/src/backends/composite.ts
```diff
@@ -24,20 +28,22 @@ import { isSandboxBackend } from "./protocol.js";
...
 
   constructor(
-    defaultBackend: BackendProtocol,
-    routes: Record<string, BackendProtocol>,
+    defaultBackend: BackendProtocol | BackendProtocolV2,
+    routes: Record<string, BackendProtocol | BackendProtocolV2>,
   ) {
-    this.default = defaultBackend;
-    this.routes = routes;
+    this.default = adaptBackendProtocol(defaultBackend);
```
hntrl: do we wanna call these types something like `AnyBackendProtocol`? then adaptBackendProtocol would be in charge of normalizing it to v2

libs/deepagents/src/backends/protocol.ts
```diff
@@ -253,10 +317,130 @@ export interface BackendProtocol {
   downloadFiles?(paths: string[]): MaybePromise<FileDownloadResponse[]>;
 }
 
+/**
+ * Updated protocol for pluggable memory backends.
```
hntrl: i wouldn't use current tense in these jsdocs

libs/deepagents/src/backends/protocol.ts
```diff
@@ -253,10 +317,130 @@ export interface BackendProtocol {
...
+}
+
 /**
  * Protocol for sandboxed backends with isolated runtime.
  * Sandboxed backends run in isolated environments (e.g., containers)
  * and communicate via defined interfaces.
+ *
+ * @deprecated Use {@link SandboxBackendProtocolV2} instead.
  */
 export interface SandboxBackendProtocol extends BackendProtocol {
```
hntrl: whats your thoughts on funneling these v1 protocols/ patterns into a dedicated `v1/` directory?

hntrl: and having `v2/` for new directory

libs/deepagents/src/backends/protocol.ts
```diff
@@ -271,18 +455,39 @@ export interface SandboxBackendProtocol extends BackendProtocol {
...
+   * Execute a command in the sandbox.
+   *
+   * @param command - Full shell command string to execute
+   * @returns ExecuteResponse with combined output, exit code, and truncation flag
+   */
+  execute(command: string): MaybePromise<ExecuteResponse>;
+
+  /** Unique identifier for the sandbox backend instance */
+  readonly id: string;
+}
```
hntrl: I'd seriously consider making this an optional `execute` method on BackendProtocolV2

libs/deepagents/src/backends/protocol.ts
```diff
@@ -271,18 +455,39 @@ export interface SandboxBackendProtocol extends BackendProtocol {
...
-  backend: BackendProtocol,
-): backend is SandboxBackendProtocol {
+  backend: BackendProtocol | BackendProtocolV2,
+): backend is SandboxBackendProtocolV2 {
   return (
-    typeof (backend as SandboxBackendProtocol).execute === "function" &&
-    typeof (backend as SandboxBackendProtocol).id === "string"
+    typeof (backend as SandboxBackendProtocolV2).execute === "function" &&
+    typeof (backend as SandboxBackendProtocolV2).id === "string"
   );
```
hntrl: this technically isn't true, right? We can put v1 BackendProtocol and it'll type to SandboxBackendProtocolV2.

also since these type casts have been awkward to work with in the past: would type `backend` to be unknown and then be exhaustive in asserting it also complies with BackendProtocol

libs/deepagents/src/backends/state.test.ts
```diff
@@ -1,6 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { StateBackend } from "./state.js";
-import type { FileData } from "./protocol.js";
+import type { FileData, FileDataV1 } from "./protocol.js";
```
hntrl, line 3: hmm is this confusing if FileDataV1 is coming from a V2 protocol?

hntrl: disregard

libs/deepagents/src/backends/protocol.ts
colifran: Ahh yeah that's a miss.

libs/deepagents/src/backends/protocol.ts
colifran: Yeah I thought about it based on our conversation but I think it is weird to have optional execute and id on BackendProtocolV2. It felt like a better design choice to keep the current structure of SandboxBackendProtocol extending the BackendProtocol. Is there anything I'm overlooking as to why we would want to do that or is this just preference?

libs/deepagents/src/backends/protocol.ts
colifran: Yeah let me take a look at how to organize that

libs/deepagents/src/backends/composite.ts
colifran: Probably cleaner then constantly adding, i.e., BackendProtocol | BackendProtocolV2 | BackendProtocolVWhatever

libs/deepagents/src/backends/protocol.ts
```diff
@@ -253,10 +317,130 @@ export interface BackendProtocol {
...
+ * Updated protocol for pluggable memory backends.
+ *
+ * Key differences from {@link BackendProtocol}:
+ * - `read()` returns {@link ReadResult} instead of a plain string
+ * - `grepRaw()` returns {@link GrepResult} instead of `GrepMatch[] | string`
+ *
+ * Existing v1 backends can be adapted to this interface using
+ * {@link adaptBackendProtocol} from utils.
+ */
+export interface BackendProtocolV2 {
```
christian-bromann: Wonder if we should move v2 primitives into a separate file, just for clear separation?

christian-bromann: Also should this protocol extend from `BackendProtocol` and just overwrite what changes?

colifran: Probably cleaner than having a bunch of repeated code - I'll take a look

libs/deepagents/src/backends/composite.ts
colifran: Updated.

libs/deepagents/src/backends/protocol.ts
colifran: Updated.

libs/deepagents/src/backends/protocol.ts
colifran: No because v1 BackendProtocols don't have execute or id on them. I think we can make the unknown change but I'm not sure we gain anything additional by being exhaustive here. The only real differentiator is execute and id

colifran: Updated backend type to be unknown.

libs/deepagents/src/backends/protocol.ts
colifran: Updated BackendProtocolV2 to extend BackendProtocol.

colifran: Added a dedicated v1 and v2 directory

libs/deepagents/src/backends/protocol.ts
colifran: Added a dedicated v1 and v2 directory

libs/deepagents/src/backends/protocol.ts
```diff
@@ -40,19 +52,70 @@ export interface GrepMatch {
...
+  /** File content as an array of lines */
   content: string[];
   /** ISO format timestamp of creation */
   created_at: string;
   /** ISO format timestamp of last modification */
   modified_at: string;
 }
 
+/**
+ * Current file data format (v2).
```
christian-bromann, line 93: Should this also live in `‎libs/deepagents/src/backends/v2/protocol.ts‎`

libs/deepagents/src/backends/composite.ts
```diff
@@ -24,20 +30,36 @@ import { isSandboxBackend } from "./protocol.js";
...
+    defaultBackend: AnyBackendProtocol,
+    routes: Record<string, AnyBackendProtocol>,
   ) {
-    this.default = defaultBackend;
-    this.routes = routes;
+    // Check if default backend is a sandbox and adapt accordingly
+    const hasExecute = typeof (defaultBackend as any).execute === "function";
+    this.default = hasExecute
+      ? adaptSandboxProtocol(defaultBackend as any)
+      : adaptBackendProtocol(defaultBackend);
```
hntrl, line 45: anywhere possible we should avoid `as any` assertions. Would it be possible to have something like `isSandboxProtocol` to make this assertion for us?

libs/deepagents/src/backends/local-shell.int.test.ts
```diff
@@ -28,7 +28,7 @@ sandboxStandardTests({
       rootDir: tmpDir,
       inheritEnv: true,
       ...options,
-    });
```
hntrl, line 31: how come

libs/deepagents/src/backends/protocol.ts
```diff
@@ -40,19 +52,100 @@ export interface GrepMatch {
...
+ * not stored on the data itself.
+ */
+export interface FileDataV2 {
+  /** File content as a single string (text or base64-encoded binary) */
+  content: string;
+  /** ISO format timestamp of creation */
+  created_at: string;
+  /** ISO format timestamp of last modification */
+  modified_at: string;
+}
```
hntrl, line 110: nit: camel case?

hntrl: also do we need to capture mimetype at this level? some providers can understand more than just `file` type

libs/deepagents/src/backends/protocol.ts
```diff
@@ -40,19 +52,100 @@ export interface GrepMatch {
...
+ * Structured result from backend read operations.
+ *
+ * Replaces the previous plain string return, giving callers a
+ * programmatic way to distinguish errors from content.
+ */
+export interface ReadResult {
+  /** Error message on failure, undefined on success */
+  error?: string;
+  /** File content as a string (text or base64-encoded binary), undefined on failure */
+  content?: string;
```
hntrl: same thing w/ mimetype

hntrl: when would this be base64 encoded? can we save the encoding time if we just always had a Uint8Array? For reference this is how we type in https://github.com/langchain-ai/langchainjs/blob/c3fcea5f288ac5264b11731d271385e9a2cee537/libs/langchain-core/src/messages/content/multimodal.ts#L51

libs/deepagents/src/backends/protocol.ts
```diff
@@ -142,150 +235,38 @@ export interface FileUploadResponse {
...
 export function isSandboxBackend(
-  backend: BackendProtocol,
-): backend is SandboxBackendProtocol {
+  backend: unknown,
+): backend is SandboxBackendProtocolV2 {
   return (
-    typeof (backend as SandboxBackendProtocol).execute === "function" &&
-    typeof (backend as SandboxBackendProtocol).id === "string"
+    typeof (backend as SandboxBackendProtocolV2).execute === "function" &&
+    typeof (backend as SandboxBackendProtocolV2).id === "string"
```
hntrl, line 275: what if backend is undefined

libs/deepagents/src/backends/utils.ts
```diff
@@ -604,3 +695,150 @@ export function formatGrepMatches(
...
+ */
+export function adaptSandboxProtocol(
+  sandbox: AnySandboxProtocol,
+): SandboxBackendProtocolV2 {
+  // First adapt the backend protocol methods to v2
+  const adapted = adaptBackendProtocol(sandbox);
+
+  // Preserve sandbox protocol properties (execute, id)
+  // Both SandboxBackendProtocol and SandboxBackendProtocolV2 have these
+  (adapted as SandboxBackendProtocolV2).execute = (cmd: string) =>
```
hntrl, line 866: is adapted not already typed as SandboxBackendPrototcolV2?

libs/deepagents/src/backends/v1/protocol.ts
```diff
@@ -0,0 +1,154 @@
...
+ * and provide a uniform interface for file operations.
+ *
+ * All file data is represented as objects with the FileData structure.
+ *
+ * Methods can return either direct values or Promises, allowing both
+ * synchronous and asynchronous implementations.
+ *
+ * @deprecated Use {@link BackendProtocolV2} instead.
+ */
+export interface BackendProtocol {
```
hntrl:
```suggestion
export interface BackendProtocolV1 {
```

then elsewhere we can do something like `export BackendProtocol extends BackendProtocolV1 {}`

more explicit naming

hntrl (commented):
on the mime type thing:
* I don't think the filename heuristic to determine what content block gets made is our best approach. E.g. in S3 I know I can attach a content-type header to a file but have an ambiguous file name

libs/deepagents/src/backends/local-shell.int.test.ts
hntrl: (as any) that is

libs/deepagents/src/backends/protocol.ts
colifran: Yeah it was just a consistency thing with the original FileData.

I thought about storing it at this level but opted against it due to being able to determine the mime type from the file extension. I don't think there is anything wrong with storing it though so we can add it.

libs/deepagents/src/backends/v1/protocol.ts
colifran: Yeah we can do that

libs/deepagents/src/backends/utils.ts
colifran: No we just know that it is a BackendProtocolV2 at this point so we need to explicitly cast it to SandboxBackendProtocolV2 here to access execute.

libs/deepagents/src/backends/composite.ts
colifran: Updated.

libs/deepagents/src/backends/protocol.ts
colifran: Improved the guard checks.

libs/deepagents/src/backends/protocol.ts
colifran: Updated.

libs/deepagents/src/backends/protocol.ts
colifran: I looked into this with Claude so please confirm but I think we would run into an issue with state and store backends. It sounds like the JsonPlusSerializer only handles Uint8Array at the top level (["bytes", obj]) but when it's nested inside the files record it goes through JSON.stringify and comes back as a plain object with numeric keys ({"0": 137, "1": 80, ...}). The replacer/reviver has cases for Set, Map, RegExp, etc. but not Uint8Array. So we would run into issues with data corruption.

libs/deepagents/src/backends/v1/protocol.ts
colifran: Done.

libs/deepagents/src/backends/local-shell.int.test.ts
colifran: Yeah good call. Fixed.

hntrl (approved):
pending a fix to JsonPlusSerializer, lgtm!

libs/deepagents/src/backends/protocol.ts
colifran: Updated to add support for `string | Uint8Array`. Added support for `Uint8Array` in the `JsonPlusSerializer` in langgraph. We need to release langgraph and bump the version here.

libs/deepagents/src/backends/protocol.ts
```diff
@@ -40,19 +64,105 @@ export interface GrepMatch {
...
+ */
+export interface FileDataV2 {
+  /** File content: string for text, Uint8Array for binary */
+  content: string | Uint8Array;
+  /** MIME type of the file (e.g. "image/png", "text/plain") */
+  mimeType: string;
+  /** ISO format timestamp of creation */
+  created_at: string;
+  /** ISO format timestamp of last modification */
+  modified_at: string;
```
hntrl, line 109: can we also make these camel case?

hntrl: or is that doing too much

libs/deepagents/src/backends/utils.ts
```diff
@@ -135,11 +225,20 @@ export function createFileData(content: string, createdAt?: string): FileData {
...
+      content: lines,
+      created_at: fileData.created_at,
+      modified_at: now,
+    };
+  }
+
   return {
-    content: lines,
+    content,
+    mimeType: (fileData as FileDataV2).mimeType,
```
hntrl: typescript doesn't intuit this?

libs/deepagents/src/backends/protocol.ts
```diff
@@ -40,19 +64,105 @@ export interface GrepMatch {
...
+  /** MIME type of the file, when available */
+  mimeType?: string;
+}
+
+/**
+ * Structured result from backend readRaw operations.
+ */
+export interface ReadRawResult {
+  /** Error message on failure, undefined on success */
+  error?: string;
```
hntrl, line 141: what was the consensus on just letting errors just be expressed through throwing an actual error object?

libs/deepagents/src/backends/protocol.ts
colifran: Again, just consistency with the existing FileData interface. Because the original properties were snake case if we make these camel case we have to do an `if isFileDataV2` everywhere to properly access createdAt vs. created_at which is really annoying. So I just chose consistency here.

libs/deepagents/src/backends/utils.ts
colifran: It does - just leftover/missed. Updated.

libs/deepagents/src/backends/protocol.ts
colifran: We do throw in some cases like when something is unrecoverable but the argument for having error here is that we can surface it to the LLM without crashing the agent. A great example is trying to have an anthropic model "read" an audio file. Audio files aren't supported by anthropic. So rather than crash we format that into something the LLM can respond with to let the user know audio files aren't supported.

PR #309, opened 2026-03-13
feat: add eval harness and initial eval suites

hntrl:
## Summary

Adds a beefier eval framework for testing agent behavior.

### `@deepagents/evals` harness (`internal/eval-harness`)

The harness decouples _what you're testing_ from _which model you're testing against_. The central abstraction is `EvalRunner`:

- **`run({ query, initialFiles? })`** — pure invocation params. The agent runs against an in-memory `StateBackend` (no containers needed — file tools operate on virtual FS in LangGraph state).
- **`extend({ systemPrompt?, tools?, subagents?, memory?, skills?, backend? })`** — returns a derived runner with agent configuration baked in. Tests customize the agent without rebuilding runners.
- **`EVAL_RUNNER` env var** selects the model. 6 runners registered out of the box: `sonnet-4-5`, `sonnet-4-5-thinking`, `opus-4-6`, `gpt-4.1`, `gpt-4.1-mini`, `o3-mini`.

Trajectory parsing walks LangGraph message arrays and groups `AIMessage` → `ToolMessage` pairs into structured steps. 4 custom vitest matchers (`toHaveAgentSteps`, `toHaveToolCallRequests`, `toHaveToolCallInStep`, `toHaveFinalTextContaining`) double as LangSmith feedback loggers — every assertion also pushes a numeric score to the experiment dashboard.

### Eval suites

| Suite | Tests | What it covers |
|---|---|---|
| `basic` | 2 | System prompt adherence, avoiding unnecessary tool calls |
| `files` | 15 | read, write, edit, ls, grep, glob, parallel I/O, deep-nested search, multi-file reasoning |
| `subagents` | 2 | `task` tool routing to named and general-purpose subagents |
| `hitl` | 3 | Human-in-the-loop interrupts — pause on configured tools, present correct review configs, resume after approval. Tests both direct agent and subagent HITL (bypasses `EvalRunner`, uses `createDeepAgent` + `MemorySaver` + `Command` directly) |
| `memory` | 6 | AGENTS.md memory injection — recall, guided naming conventions, code style influence, multiple sources, graceful missing files, avoiding redundant reads |
| `skills` | 6 | SKILL.md discovery and usage — read by description match, read by name, combine two skills in parallel, edit skills (with and without prior read), resolve correct skill across multiple source paths |
| `tool-usage-relational` | 18 | Ported from Python `test_tool_usage_relational.py`. Relational data with fake users/locations/foods connected by IDs. Tests sequential tool chaining (1→2→3→4 hops) and parallel fan-out (e.g. resolve 3 food IDs in parallel after a sequential lookup chain) |

### Skills

- **`eval-creator`** (`.agents/skills/eval-creator/`) — comprehensive guide for scaffolding new eval suites. Covers inline, fixture, external dataset, and LangSmith dataset patterns; sandbox-backed evals; LLM-as-judge scoring; published benchmark implementation.
- **`langsmith-trace`** (`.agents/skills/langsmith-trace/`) — replaces the trace skill from the deleted langsmith-skills subtree, installed via `skills-lock.json`.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -176,12 +176,6 @@ importers:
       '@deepagents/evals':
         specifier: workspace:*
         version: link:../../internal/eval-harness
-      '@langchain/anthropic':
-        specifier: ^1.3.23
-        version: 1.3.23(@langchain/core@1.1.32(@opentelemetry/api@1.9.0)(@opentelemetry/exporter-trace-otlp-proto@0.207.0(@opentelemetry/api@1.9.0))(@opentelemetry/sdk-trace-base@2.5.1(@opentelemetry/api@1.9.0))(openai@6.29.0(ws@8.19.0)(zod@4.3.6))(ws@8.19.0))
-      '@langchain/quickjs':
-        specifier: workspace:*
-        version: link:../../libs/providers/quickjs
       deepagents:
         specifier: workspace:*
         version: link:../../libs/deepagents


evals/README.md
```diff
@@ -0,0 +1,156 @@
...
+// Seed files
+const result = await runner.run({
+  query: "Read /data.txt",
+  initialFiles: { "/data.txt": "hello world" },
+});
+```
+
+### Custom matchers
+
+The harness provides vitest matchers that also log LangSmith feedback:
```
maahir30, line 146: are we sure that we want efficiency metrics to induce failures? in the Python version, `agent_steps` `tool_call_requests`  and `tool_call` checks are `EfficiencyAssertions` that never fail the test. They're logged to LangSmith for tracking but don't block CI. I believe this is intentional bc a model that takes 3 steps instead of 2 still answered correctly. We would call that an efficiency regression, not an explicit failure. This could cause a problem bc swapping models will likely cause mass "failures" even if correctness is fine. But I can see a case for the other side too just wanted to flag.

PR #317, opened 2026-03-17
fix(deepagents, quickjs): read store from runtime/config.store instead of config.configurable

hntrl:
## Summary

Fix store propagation in filesystem and QuickJS middleware to use properly typed access paths instead of reaching into internal configurable fields.

### `deepagents` — `libs/deepagents/src/middleware/fs.ts`

The filesystem middleware was accessing the store via `request.config.store` with `@ts-expect-error` suppressions. This worked by accident because LangGraph happened to attach the store there, but it wasn't part of the typed contract. Updated both call sites in `createFilesystemMiddleware` to read from `request.runtime.store` instead, which is the intended propagation path. This removes two `@ts-expect-error` comments.

### `@langchain/quickjs` — `libs/providers/quickjs/src/middleware.ts`

The QuickJS middleware was reading the store from `config.configurable.__pregel_store`, an internal/private LangGraph field. Updated to use `config.store`, which is the public API exposed by `LangGraphRunnableConfig`. Added the `LangGraphRunnableConfig` type import and annotated the tool callback parameter for proper typing.

### Test updates

Updated tests in `composite.test.ts`, `state.test.ts`, and `store.test.ts` to pass the store via `runtime: { store }` instead of `config: { store }`, matching the new propagation path.

### DIFF
--- a/libs/providers/quickjs/src/middleware.ts
+++ b/libs/providers/quickjs/src/middleware.ts
@@ -47,7 +47,10 @@ import {
 import type * as _zodTypes from "@langchain/core/utils/types";
 import type * as _zodMeta from "@langchain/langgraph/zod";
 import type * as _messages from "@langchain/core/messages";
-import { getCurrentTaskInput } from "@langchain/langgraph";
+import {
+  getCurrentTaskInput,
+  LangGraphRunnableConfig,
+} from "@langchain/langgraph";
 
 /**
  * Backend-provided tools excluded from PTC by default.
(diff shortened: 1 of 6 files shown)


libs/providers/quickjs/src/middleware.ts
```diff
@@ -224,12 +227,12 @@ export function createQuickJSMiddleware(
   }
 
   const jsEvalTool = tool(
-    async (input, config) => {
+    async (input, config: LangGraphRunnableConfig) => {
```
christian-bromann, line 230: Shouldn't this be `ToolRuntime`?

PR #322, opened 2026-03-18
feat(node-vfs): enhance VfsSandbox to support absolute paths execute

hnustwjj:
# fix(node-vfs): rewrite absolute VFS paths in execute() commands

## Summary

- Fix `VfsSandbox.execute()` to support absolute paths (e.g. `node /src/index.js`) by rewriting them to the temp directory path before shell execution
- Add integration tests covering absolute path command execution for single-file and multi-file Node.js projects

## Problem

When users define `initialFiles` with absolute paths like `/src/index.js`, the VFS correctly stores and syncs files to a temp directory for execution. However, commands referencing those files with absolute paths fail:

```typescript
const sandbox = await VfsSandbox.create({
  initialFiles: { "/src/index.js": "console.log('Hello')" },
});

await sandbox.execute("node src/index.js");   // ✅ works (relative → resolves via cwd)
await sandbox.execute("node /src/index.js");  // ❌ fails (absolute → resolves against host root)
```

The `#normalizeExecPath` method already handles leading `/` for `read()`, `ls()`, `grep()`, `glob()`, but `execute()` passes the raw command string directly to `/bin/bash -c` without any path translation.

## Solution

Add a `#rewriteVfsPaths(command, execDir)` private method that:

1. Reads the top-level entry names from the VFS workspace (e.g. `src`, `data.json`)
2. For each entry, uses a regex to match `/<name>` at path-token boundaries in the command
3. Replaces matched absolute paths with `<tempDir>/<name>`

The regex uses negative lookbehind (`(?<![\w/.-])`) and positive lookahead (`(?=/|[\s"';|&><!]|$)`) to ensure only standalone absolute path tokens are rewritten, leaving system paths like `/bin/bash` or `/usr/local/bin/node` untouched (since `bin`, `usr` are not VFS entries).

## Test plan

- [x] New test: `should run a multi-file Node.js project with absolute paths` — verifies `node /src/index.js` with cross-file `require()`
- [x] New test: `should access initialFiles with leading slash via execute() using absolute path` — verifies single-file absolute path execution
- [x] New test: `should run a multi-file Node.js project with absolute paths from README (absolute execute)` — mirrors README example with absolute path
- [x] All 28 unit tests pass
- [x] All 125 integration tests pass (including 92 standard tests)

### DIFF
--- a/libs/providers/node-vfs/src/sandbox.ts
+++ b/libs/providers/node-vfs/src/sandbox.ts
@@ -308,10 +308,7 @@ export class VfsSandbox extends BaseSandbox {
       //   inside longer paths like "foo/src" or "../src")
       // - Followed by /, whitespace, quotes, shell operators, or end
       result = result.replace(
-        new RegExp(
-          `(?<![\\w/.-])/${escaped}(?=/|[\\s"';|&><!\\])]|$)`,
-          "gm",
-        ),
+        new RegExp(`(?<![\\w/.-])/${escaped}(?=/|[\\s"';|&><!\\])]|$)`, "gm"),
         `${execDir}/${name}`,
       );
     }


christian-bromann (approved):
LGTM 👍

hnustwjj:
@christian-bromann thank you for your approve. i'll be appreciate if  you could help me to review another [pr](https://github.com/langchain-ai/langchainjs/pull/10481)?

PR #323, opened 2026-03-18
feat(sdk): add async subagent middleware for remote LangGraph servers 

colifran:
### Summary

  Adds `AsyncSubAgentMiddleware` for launching background jobs on remote LangGraph deployments via the LangGraph SDK. This is the JS port of the Python `AsyncSubAgentMiddleware`.

  The middleware provides five tools:
  - `start_async_task` — creates a thread + run on a remote server, returns a job ID immediately
  - `check_async_task` — fetches run status and extracts the result on completion
  - `update_async_task` — sends follow-up instructions by creating a new run on the same thread (with `multitaskStrategy: "interrupt"`)
  - `cancel_async_task` — cancels a running job
  - `list_async_tasks` — lists all tracked jobs with live statuses fetched in parallel

  Job state is persisted in the `asyncSubAgentJobs` state channel via a `ReducedValue` reducer, so jobs survive context compaction and summarization.

  Integrated into `createDeepAgent` via the `asyncSubagents` parameter.

  ### Tests

  77 unit tests covering:
  - State reducer merge logic and immutability
  - `ClientCache` — caching, key deduplication, default `x-auth-scheme` header injection
  - All 5 tools — happy paths, unknown job IDs, SDK errors, graceful degradation (e.g. `threads.getState` failure on success still returns a result)
  - `buildCheckResult` — message content extraction, non-object fallback, empty messages fallback, error status
  - `buildListTool` — status filtering, terminal status SDK call skipping, live status updates, parallel fetching
  - `createAsyncSubAgentMiddleware` — validation (empty list, duplicate names), tool wiring, system prompt injection, independent instances

### DIFF
--- a/libs/deepagents/src/middleware/async_subagents.ts
+++ b/libs/deepagents/src/middleware/async_subagents.ts
@@ -424,7 +424,7 @@ export class ClientCache {
  * Creates a thread on the remote server, starts a run, and returns a
  * `Command` that persists the new task in state.
  */
-export function buildLaunchTool(
+export function buildStartTool(
   agentMap: Record<string, AsyncSubAgent>,
   clients: ClientCache,
   toolDescription: string,
@@ -834,7 +834,7 @@ export function createAsyncSubAgentMiddleware(
   );
 
   const tools = [
-    buildLaunchTool(agentMap, clients, launchDescription),
+    buildStartTool(agentMap, clients, launchDescription),
     buildCheckTool(clients),
     buildUpdateTool(agentMap, clients),
     buildCancelTool(clients),
--- a/libs/deepagents/src/middleware/async_subagents.test.ts
+++ b/libs/deepagents/src/middleware/async_subagents.test.ts
@@ -6,7 +6,7 @@ import type { ToolRuntime } from "langchain";
 
 import {
   asyncTasksReducer,
-  buildLaunchTool,
+  buildStartTool,
   buildCheckTool,
   buildUpdateTool,
   buildCancelTool,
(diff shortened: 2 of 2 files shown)


libs/deepagents/src/middleware/async_subagents.ts
```diff
@@ -0,0 +1,833 @@
...
+export function buildUpdateTool(
+  agentMap: Record<string, AsyncSubAgent>,
+  clients: ClientCache,
+) {
+  return tool(
+    async (
+      input: { jobId: string; message: string },
+      config,
+    ): Promise<Command | string> => {
+      const state = getCurrentTaskInput<AsyncSubAgentState>();
```
christian-bromann: Doesn't `config.state` yield the agent stage? I don't think we need to use `getCurrentTaskInput` which looks very nebulous how it retrieves the state value.

colifran: You might be right. Let me take a look.

colifran: Doesn't look like there is a state property on config.

colifran: Looks like a similar pattern is used in for subagent: https://github.com/langchain-ai/deepagentsjs/blob/d398e518d0b1452c6e59711073d070a9d7dd1740/libs/deepagents/src/middleware/subagents.ts#L596

christian-bromann: > Doesn't look like there is a state property on config.

It should be:

```ts
import { type ToolRuntime, tool } from "langchain";

interface MyState {
  messages: any
}

const t = tool((params: any, runtime: ToolRuntime<MyState>) => {
  console.log(runtime.state)
}, {
  ...
})
```

You have to explicitly type it as `ToolRuntime`.

colifran: Ah I see. Thanks for the example. Updated!

PR #330, opened 2026-03-19
fix(deepagents): throw on built-in tool collision

maahir30:
- `createDeepAgent` now throws at construction time if any user-supplied tool name collides with a built-in tool (`ls`, `read_file`, `write_file`, `edit_file`, `glob`, `grep`, `execute`, `task`, `write_todos`). Previously, colliding tools silently shadowed the built-in, causing cryptic schema-validation errors at runtime.
- Adds `FILESYSTEM_TOOL_NAMES` constant in `fs.ts` as the single source of truth for filesystem middleware tool names.

### DIFF
--- a/libs/deepagents/src/middleware/fs.ts
+++ b/libs/deepagents/src/middleware/fs.ts
@@ -907,31 +907,36 @@ export function createFilesystemMiddleware(
 
   const baseSystemPrompt = customSystemPrompt || FILESYSTEM_SYSTEM_PROMPT;
 
-  // All tools including execute (execute will be filtered at runtime if backend doesn't support it)
-  const allTools = [
-    createLsTool(backend, {
+  /**
+   * All tools including execute
+   * (execute will be filtered at runtime if backend doesn't support it)
+   */
+  type FilesystemToolName = (typeof FILESYSTEM_TOOL_NAMES)[number];
+  const allToolsByName = {
+    ls: createLsTool(backend, {
       customDescription: customToolDescriptions?.ls,
     }),
-    createReadFileTool(backend, {
+    read_file: createReadFileTool(backend, {
       customDescription: customToolDescriptions?.read_file,
       toolTokenLimitBeforeEvict,
     }),
-    createWriteFileTool(backend, {
+    write_file: createWriteFileTool(backend, {
       customDescription: customToolDescriptions?.write_file,
     }),
-    createEditFileTool(backend, {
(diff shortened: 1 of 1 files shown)


libs/deepagents/src/agent.ts
```diff
@@ -130,6 +137,20 @@ export function createDeepAgent<
...
+    .filter(
+      (n): n is string => typeof n === "string" && BUILTIN_TOOL_NAMES.has(n),
+    );
+
+  if (collidingTools.length > 0) {
+    throw new Error(
+      `[deepagents] Tool name(s) [${collidingTools.join(", ")}] ` +
+        `conflict with built-in tools. ` +
+        `Rename your custom tools to avoid this.`,
+    );
```
hntrl: think we should stand up error classes like we do in google instead of throwing an opaque error. E.g. ConfigurationError or something like that

libs/deepagents/src/middleware/fs.ts
```diff
@@ -62,6 +62,21 @@ import {
...
+ */
+export const FILESYSTEM_TOOL_NAMES = [
+  "ls",
+  "read_file",
+  "write_file",
+  "edit_file",
+  "glob",
+  "grep",
+  "execute",
+] as const;
```
christian-bromann, line 78: Can we ensure that this always matches the tools we pass in, e.g.:
```ts
  type FilesystemToolName = (typeof FILESYSTEM_TOOL_NAMES)[number];
  const allToolsByName: Record<FilesystemToolName, unknown> = {
    ls: createLsTool(backend, {
      customDescription: customToolDescriptions?.ls,
    }),
    read_file: createReadFileTool(backend, {
      customDescription: customToolDescriptions?.read_file,
      toolTokenLimitBeforeEvict,
    }),
    write_file: createWriteFileTool(backend, {
      customDescription: customToolDescriptions?.write_file,
    }),
    edit_file: createEditFileTool(backend, {
      customDescription: customToolDescriptions?.edit_file,
    }),
    glob: createGlobTool(backend, {
      customDescription: customToolDescriptions?.glob,
    }),
    grep: createGrepTool(backend, {
      customDescription: customToolDescriptions?.grep,
    }),
    execute: createExecuteTool(backend, {
      customDescription: customToolDescriptions?.execute,
    }),
  };
  const allTools = Object.values(allToolsByName);
```

This ensures we update the list when adding a new tool.

PR #334, opened 2026-03-20
feat(deepagents): add completion notifier middleware for async subagents

hntrl:
## Summary

Port of [langchain-ai/deepagents#2119](https://github.com/langchain-ai/deepagents/pull/2119) to TypeScript.

Adds a `createCompletionNotifierMiddleware` that async subagents can use to proactively notify their supervisor when they complete or error -- closing the gap where the supervisor only learns about completion when someone calls `check_async_task`.

## Why

The async subagent protocol is fire-and-forget: the supervisor launches a task and only discovers its outcome when it (or the user) polls via `check_async_task`. This means the supervisor can't proactively relay results -- the user has to ask. The completion notifier solves this by having the subagent push a notification to the supervisor's thread when it finishes.

## Architecture

```
Supervisor                    Subagent
    |                            |
    |--- start_async_task -----> |
    |<-- task_id (immediately) - |
    |                            |  (working...)
    |                            |  (done!)
    |                            |
    |<-- runs.create(            |
    |      supervisor_thread,    |
    |      "completed: ...")     |
    |                            |
    |  (wakes up, sees result)   |
```

This is an opt-in middleware added to the **subagent's** stack, not the supervisor's. The notifier calls `runs.create()` on the supervisor's thread via the `@langchain/langgraph-sdk` Client, which queues a new run that wakes the supervisor.

**Parent context propagation:** `parent_thread_id` is injected into the subagent's input state by the supervisor's `start_async_task` tool (from the Python PR). The `parentGraphId` (supervisor's graph/assistant ID) is passed as a constructor argument since it's static configuration known at deployment time.

### New: `createCompletionNotifierMiddleware` (`libs/deepagents/src/middleware/completion_notifier.ts`)
- `afterAgent` hook sends a completion notification with result summary (truncated to 500 chars)
- `wrapModelCall` hook catches errors and sends an error notification before re-raising
- State schema adds `parent_thread_id` to the subagent's state
- Notifies only once per run (guards against duplicates)
- Silently no-ops if parent context is missing (subagent launched without a supervisor)
- Supports `url` and `headers` options for remote supervisors

### New: `completion_notifier.test.ts` — 22 unit tests
- Tests for `extractLastMessage`, `notifyParent`, `afterAgent`, `wrapModelCall`
- Covers edge cases: missing parent context, duplicate notification guard, error swallowing

### Modified: `package.json`
- Added `@langchain/langgraph-sdk` as a required dependency

### Exports
- `createCompletionNotifierMiddleware` and `CompletionNotifierOptions` exported from `deepagents` top-level and `middleware/index.ts`

## Usage

```typescript
import { createCompletionNotifierMiddleware, createDeepAgent } from "deepagents";

// Same deployment (ASGI transport):
const notifier = createCompletionNotifierMiddleware({
  parentGraphId: "supervisor",
});

// Remote deployment:
const notifier = createCompletionNotifierMiddleware({
  parentGraphId: "supervisor",
  url: "https://supervisor.langsmith.dev",
});

  model: "claude-sonnet-4-5-20250929",
  middleware: [notifier],
});
```

## Divergence from Python

The Python implementation uses `langgraph-sdk`'s `get_client()` with ASGI transport for same-deployment communication. The JS version uses `@langchain/langgraph-sdk`'s `Client` class with a default localhost URL (`http://localhost:8123`) when no URL is provided. The `parentGraphId` parameter matches the Python `parent_graph_id` constructor arg.

### DIFF
--- a/.changeset/clear-years-end.md
+++ b/.changeset/clear-years-end.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+feat(deepagents): add completion notifier middleware for async subagents
--- a/.changeset/fresh-ways-cross.md
+++ b/.changeset/fresh-ways-cross.md
@@ -1,5 +1,0 @@
----
-"deepagents": patch
----
-
-fix(deepagents): remove orphaned ToolMessages for Gemini compatibility
--- a/libs/deepagents/package.json
+++ b/libs/deepagents/package.json
@@ -40,7 +40,7 @@
   "dependencies": {
     "@langchain/core": "^1.1.33",
     "@langchain/langgraph": "^1.1.4",
-    "@langchain/langgraph-sdk": "^1.7.3",
+    "@langchain/langgraph-sdk": "^1.8.0",
     "fast-glob": "^3.3.3",
     "langchain": "1.2.36",
     "micromatch": "^4.0.8",
--- a/libs/deepagents/src/index.ts
+++ b/libs/deepagents/src/index.ts
@@ -60,8 +60,9 @@ export {
   DEFAULT_GENERAL_PURPOSE_DESCRIPTION,
   DEFAULT_SUBAGENT_PROMPT,
   TASK_SYSTEM_PROMPT,
-  // Async subagent constants
-  ASYNC_TASK_SYSTEM_PROMPT,
+  // Completion notifier middleware for async subagents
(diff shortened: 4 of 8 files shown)


colifran (approved):
Nice!

PR #336, opened 2026-03-22
fix(deepagents): use `crypto.randomUUID()` instead of uuid

christian-bromann:
`uuid` is ESM only. Let's use `crypto.randomUUID()` instead.

### DIFF
--- a/libs/deepagents/src/middleware/fs.int.test.ts
+++ b/libs/deepagents/src/middleware/fs.int.test.ts
@@ -13,8 +13,6 @@ import {
   StoreBackend,
   CompositeBackend,
 } from "../backends/index.js";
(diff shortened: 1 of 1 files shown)


libs/deepagents/src/middleware/fs.int.test.ts
```diff
@@ -13,7 +13,9 @@ import {
   StoreBackend,
   CompositeBackend,
 } from "../backends/index.js";
-import { v4 as uuidv4 } from "uuid";
+import { fileDataToString } from "../backends/utils.js";
+import type { FileData } from "../backends/protocol.js";
```
christian-bromann:
```suggestion
```

PR #427, opened 2026-04-05
feat(providers): add @langchain/wasmsh — in-process shell + Python sandbox

johannhartmann:
## Summary

Adds `@langchain/wasmsh`, a sandbox provider that runs a full Bash-compatible
shell (88 utilities including grep, sed, awk, jq, curl) and Python 3.13 with
pip — entirely in-process via WebAssembly. No containers, no cloud services,
no API keys.

- **Node.js**: `WasmshSandbox.createNode()` — spawns a local host process
- **Browser**: `WasmshSandbox.createBrowserWorker()` — runs in a Web Worker

Backed by [wasmsh](https://github.com/mayflower/wasmsh) and
[Pyodide](https://pyodide.org/), the shell and Python share a virtual
filesystem. Agents get `execute`, `read_file`, `write_file`, `edit_file`,
`ls`, `grep`, `glob` — same tools as remote sandboxes, zero infrastructure.

### What's included

| Commit | Change |
|--------|--------|
| `feat: expose filesystemOptions in createDeepAgent` | Allows tuning token eviction thresholds per agent |
| `feat: add @langchain/wasmsh sandbox provider` | Main provider package with 32 unit + 97 integration tests |
| `feat: add browser build and LLM agent integration tests` | Browser entry for deepagents core, Playwright e2e, agent tests |
| `fix: browser subagent state handling` | Use `runtime.state` instead of `getCurrentTaskInput()` in browser environments |
| `chore: add changeset` | Changeset for release |

### Why not just use containers?

Containers are the right choice for untrusted code or system-level operations.
Wasmsh is for the common case where agents need a filesystem, a shell, and
Python — and you don't want to spin up infrastructure for it. Tests run in
<1s, CI needs no secrets, and it works in the browser.

## Test plan

- [x] 32 unit tests (mocked session, all sandbox operations)
- [x] 97 integration tests (standard test suite, LLM agent tests)
- [x] 821 core tests pass (no regressions, typecheck clean)
- [x] 125 node-vfs tests pass (no regressions to other providers)
- [x] Playwright browser e2e tests
- [x] `pnpm format:check` clean
- [x] `pnpm lint` clean (0 errors)
- [x] `pnpm build` succeeds

### DIFF
--- a/.changeset/add-wasmsh-sandbox.md
+++ b/.changeset/add-wasmsh-sandbox.md
@@ -1,0 +1,17 @@
+---
+"deepagents": minor
+"@langchain/wasmsh": minor
+---
+
+feat: add @langchain/wasmsh sandbox provider
+
+Adds a new sandbox provider backed by wasmsh, a Bash-compatible shell
+runtime compiled to WebAssembly. Runs entirely in-process with no
+containers or remote services — works in both Node.js and browser
+Web Workers.
+
+- New package `@langchain/wasmsh` with `WasmshSandbox.createNode()` and
+  `WasmshSandbox.createBrowserWorker()` factory methods
+- Browser build entry for deepagents core (`index.browser.js`)
+- `filesystemOptions` parameter on `createDeepAgent` for middleware tuning
+- Browser-compatible subagent state handling
--- a/examples/sandbox/wasmsh-browser-sandbox.ts
+++ b/examples/sandbox/wasmsh-browser-sandbox.ts
@@ -1,0 +1,19 @@
+/* eslint-disable no-console */
+import { WasmshSandbox } from "@langchain/wasmsh";
+
+async function main() {
+  const sandbox = await WasmshSandbox.createBrowserWorker({
+    assetBaseUrl: "/node_modules/wasmsh-pyodide/assets",
+  });
+
+  try {
+    const result = await sandbox.execute(
+      "python3 -c \"print('hello from browser worker')\"",
+    );
+    console.log(result.output.trim());
+  } finally {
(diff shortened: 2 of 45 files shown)


johannhartmann:
Thanks for the catch — this is a real escape. Fixed in 799d461 on this branch.

**Fix:**
`#scope` now resolves the joined path via `posix.resolve` and rejects if the result no longer sits at-or-below the namespace prefix. `#unscope` applies the matching containment check on inbound paths so an upstream bug elsewhere can't leak non-namespaced paths back to callers. The prefix match is anchored with a trailing slash so a sibling whose name shares the prefix (`/memstore` vs `/mem`) is also rejected — your recommendation list covered that case implicitly via `posix.resolve`, but I made it explicit so the guard reads top-to-bottom.

The rejection surfaces as `WasmshNamespaceEscapeError` (exported).

**Tests added** (8 cases in `filesystem-backend.test.ts`):

- direct `..` escape rejected on every protocol method (read/write/edit/ls/glob/grep/upload/download)
- multi-segment `../../skills/x` payload
- interior `/x/../../etc/passwd` that resolves outside
- sibling-prefix attack (`/../memstore/x` against `/mem`)
- benign `./sub` and interior `..` that stays inside (still allowed)
- upstream-leak guard on `#unscope` (simulated sandbox returns `/etc/passwd`)
- no-namespace passthrough (containment doesn't apply when no namespace configured)

The same bug was in the Python adapter (`langchain_wasmsh.WasmshFilesystemBackend`) and is fixed identically in mayflower/wasmsh@29a6c2f, with the same 6 regression tests on the Python side.

PR #448, opened 2026-04-10
fix: follow symlinks in sandbox find commands

### LINKED ISSUE #447: Bug: sandbox ls/glob/grep tools don't follow symlinks
## Description

The `BaseSandbox` class in `libs/deepagents/src/backends/sandbox.ts` uses `find` without the `-L` flag in three helper functions:

- `buildLsCommand` — `find ${quotedPath} -maxdepth 1 ...`
- `buildFindCommand` — `find ${quotedPath} ...`
- `buildGrepCommand` — `find ${searchPathQuoted} -type f -name ...`

Without `-L`, `find` does not dereference symbolic links. This means that when a path is itself a symlink (or contains symlinked directories), `find` skips the target entirely and returns no results.

## Impact

On **Modal** (and potentially other sandbox providers), volumes are mounted as symlinks. As a result, all three sandbox tools (`ls`, `glob`, `grep`) return empty results when pointed at symlinked paths, making the agent unable to explore or search the filesystem.

## Reproduction Steps

1. Set up a Modal sandbox with a volume mount (e.g., `/vol` -> symlinked NFS mount)
2. Use the `ls` tool to list files in `/vol`
3. Observe: returns empty results
4. Run `find -L /vol -maxdepth 1` manually inside the sandbox — returns the expected files

## Expected Behavior

`ls`, `glob`, and `grep` tools should follow symlinks and return results from symlinked directories, matching what a user would see with standard `ls` or `find -L`.

## Actual Behavior

All three tools return empty results for any path that is a symlink.

## Proposed Fix

Add the `-L` flag to all three `find` invocations in `buildLsCommand`, `buildFindCommand`, and `buildGrepCommand`. The `-L` flag causes `find` to follow symbolic links, which is the expected behavior for filesystem exploration tools.

ItayCoCo:
## Summary

- Add `-L` flag to `find` in `buildLsCommand`, `buildFindCommand`, and `buildGrepCommand` so that symlinks are followed
- Fixes #447

## Problem

`BaseSandbox` helper functions use `find` without `-L`, which means symbolic links are not dereferenced. On Modal (and potentially other providers), volumes are mounted as symlinks, so the `ls`, `glob`, and `grep` sandbox tools all return empty results when pointed at symlinked paths.

## Fix

The `-L` flag tells `find` to follow symbolic links. This is a one-line change in each of the three helper functions:

```diff
- find ${quotedPath} -maxdepth 1 ...
+ find -L ${quotedPath} -maxdepth 1 ...
```

This matches user expectations — filesystem exploration tools should transparently follow symlinks, just like `ls` and `find -L` do in a normal shell.

## Test plan

- [ ] Verify `ls` tool returns files when the target path is a symlink
- [ ] Verify `glob` tool traverses symlinked directories
- [ ] Verify `grep` tool searches inside symlinked directories
- [ ] Confirm no regression on non-symlinked paths (behavior is identical when no symlinks are present)

### DIFF
--- a/.changeset/fix-follow-symlinks.md
+++ b/.changeset/fix-follow-symlinks.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+fix: follow symlinks in sandbox find commands by adding -L flag to find invocations in buildLsCommand, buildFindCommand, and buildGrepCommand
--- a/libs/deepagents/src/backends/sandbox.test.ts
+++ b/libs/deepagents/src/backends/sandbox.test.ts
@@ -179,6 +179,8 @@ describe("BaseSandbox", () => {
       expect(sandbox.executedCommands[0]).toContain("-maxdepth 1");
     });
 
+    // Symlinked volumes (e.g. Modal) are invisible without -L.
+    // See: https://github.com/langchain-ai/deepagentsjs/issues/447
     it("should include -L flag to follow symlinks", async () => {
       const sandbox = new MockSandbox();
       sandbox.addFile("/test.txt", "content");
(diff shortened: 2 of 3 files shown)


christian-bromann (commented):
Thanks for the contribution. Can we have a unit test for this? I think it well help prevent us from accidentally removing the behavior.

ItayCoCo:
Added!

hntrl (approved):

PR #453, opened 2026-04-13
swarm refactor

colifran:


libs/deepagents/src/middleware/swarm.ts
```diff
@@ -0,0 +1,260 @@
...
+        // File written via execute — read from filesystem directly
+        try {
+          content = readFileSync(tasksPath, "utf-8");
+        } catch {
+          return (
+            `Failed to read tasks file at "${tasksPath}". ` +
+            `Ensure the generation script writes the file to this exact path and try again.`
+          );
+        }
+      }
```
ccurme, line 168: why do we need to fall back to this instead of going through backend?

libs/deepagents/src/swarm/executor.ts
```diff
@@ -0,0 +1,283 @@
...
+ * Write the results table to a unique run directory.
+ *
+ * @returns The run directory path (e.g. `swarm_runs/<uuid>`)
+ */
+async function writeResults(
+  backend: BackendProtocolV2,
+  results: SwarmTaskResult[],
+): Promise<string> {
+  const runId = randomUUID();
+  const resultsDir = `swarm_runs/${runId}`;
```
ccurme, line 173: do we need a leading `/` here

colifran: Nice. Good catch.

PR #463, opened 2026-04-14
fix: skills middleware discovery for scoped directories - Fixes #387

thushanthbengre22-dev:
## Description

Fixes skills middleware skipping `SKILL.md` discovery when explicitly scoped skill directories are provided.

## Issue

Closes #387

## Problem

When `skills` is configured with a specific skill directory path like `"skills/domainA/"`, the `listSkillsFromBackend` function fails to discover `SKILL.md` at that location.

**Root cause:**  
The function calls `ls("skills/domainA/")`, which returns the **contents** of that directory (files like `SKILL.md`). The existing loop only processes subdirectories and skips files, so `SKILL.md` is never discovered.

**Result:**
- `skills_locations` shows `"skills/domainA/"` in the system prompt
- `skills_list` shows `"none"`
- Internally inconsistent state (noted at line 715)

## Solution

Before the subdirectory iteration loop, check if `SKILL.md` exists as a direct child of the source directory. If found:
1. Parse the skill metadata directly
2. Add to skills list
3. Return early

This enables scoped skill directories while maintaining backward compatibility with the existing parent directory behavior.

## Changes

**Modified:** `libs/deepagents/src/middleware/skills.ts`
- Added check for `SKILL.md` as direct child before subdirectory loop
- Reads and parses skill if found at source path
- Returns early to avoid unnecessary processing

**Modified:** `libs/deepagents/src/middleware/skills.test.ts`
- Added regression test reproducing the bug scenario
- Verifies `SKILL.md` is discovered when source is explicit skill directory

## Testing

**New test case:**
```javascript
sources: ["/skills/domainA/"]
// With SKILL.md at /skills/domainA/SKILL.md
```

**Before fix:** `skillsMetadata` was empty ❌  
**After fix:** `skillsMetadata` contains the discovered skill ✅

**Backward compatibility verified:**
- ✅ Parent directories still work: `sources: ["skills/"]`
- ✅ Multiple scoped directories: `sources: ["skills/domainA/", "skills/domainB/"]`
- ✅ All existing tests pass

## Checklist

- [x] Code follows project style guidelines
- [x] Added test coverage for the fix
- [x] All tests pass locally
- [x] Backward compatible - no breaking changes

### DIFF
--- a/.changeset/clean-onions-reply.md
+++ b/.changeset/clean-onions-reply.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+Fix skills middleware discovery for scoped directories (#387)
--- a/libs/deepagents/src/middleware/skills.test.ts
+++ b/libs/deepagents/src/middleware/skills.test.ts
@@ -769,9 +769,7 @@ description: [invalid yaml syntax: unclosed bracket
       // explicitly scoped source directory.
       expect(result?.skillsMetadata).toHaveLength(1);
       expect(result?.skillsMetadata[0].name).toBe("web-research");
-      expect(result?.skillsMetadata[0].path).toBe(
-        "/skills/domainA/SKILL.md",
-      );
+      expect(result?.skillsMetadata[0].path).toBe("/skills/domainA/SKILL.md");
     });
   });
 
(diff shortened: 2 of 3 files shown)


hntrl (commented):
Closing in favor of #242

PR #464, opened 2026-04-14
fix(deepagents): follow symlinks in sandbox find commands

### LINKED ISSUE #447: Bug: sandbox ls/glob/grep tools don't follow symlinks

hntrl:
## Summary

Fixes #447.

Sandbox filesystem discovery commands were not following symlinks, causing `ls`, `glob`, and `grep` (with `glob`) to return empty results for symlinked directories (for example, provider mounts exposed via symlinks). This change updates the `BaseSandbox` command builders to use `find -L` so symlink targets are traversed consistently.

### `deepagents` (`libs/deepagents/src/backends/sandbox.ts`)

- Updated `buildLsCommand` to use `find -L`.
- Updated `buildFindCommand` (used by `glob`) to use `find -L`.
- Updated glob-filtered `buildGrepCommand` path to use `find -L` before `grep`.

### Tests (`libs/deepagents/src/backends/sandbox.test.ts`)

- Added assertions that generated `ls` and `glob` commands include `find -L`.
- Added coverage for `grep(..., glob=...)` to verify symlink-following command generation.

### DIFF
--- a/libs/deepagents/src/backends/sandbox.test.ts
+++ b/libs/deepagents/src/backends/sandbox.test.ts
@@ -175,6 +175,7 @@ describe("BaseSandbox", () => {
       await sandbox.ls("/");
       expect(sandbox.executedCommands.length).toBeGreaterThan(0);
       expect(sandbox.executedCommands[0]).toContain("find");
+      expect(sandbox.executedCommands[0]).toContain("find -L");
       expect(sandbox.executedCommands[0]).toContain("stat");
       expect(sandbox.executedCommands[0]).toContain("-maxdepth 1");
     });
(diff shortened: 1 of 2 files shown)


christian-bromann:
How does it differ from #448?

hntrl:
ah I missed that

PR #472, opened 2026-04-17
feat(deepagents): implement swarm as a foreign function in quickjs middleware to enable parallel task fan-out

colifran:


libs/providers/quickjs/src/middleware.ts
```diff
@@ -62,13 +68,20 @@ export const DEFAULT_PTC_EXCLUDED_TOOLS = [
   "execute",
 ] as const;
 
-const REPL_SYSTEM_PROMPT = dedent`
+function buildReplSystemPrompt(hasSwarm: boolean): string {
+  const largeFileDirective = hasSwarm
+    ? `**Large file processing**: Before doing anything else, check the size of your input files. If any input file exceeds ~50KB, you must decompose the work — use \`js_eval\` with \`swarm()\` to split the data and dispatch parallel tasks. Do not attempt to process large files inline. See "Parallel fan-out" below.`
```
ccurme: do we need sandboxes to measure file size in kb? will quickJS do for us?

PR #483, opened 2026-04-22
feat(deepagents): expose summarization options on createDeepAgent

ariran5:
Add optional `summarization` parameter (DeepAgentSummarizationOptions) so callers can pass trigger, keep, trimTokensToSummarize, and other SummarizationMiddleware settings without duplicating middleware wiring. Options are forwarded to createSummarizationMiddleware for the root agent and all built-in subagent stacks; backend always comes from the agent.


hntrl (commented):
See https://github.com/langchain-ai/deepagentsjs/issues/484#issuecomment-4606608373

PR #486, opened 2026-04-23
feat(quickjs): remove built-in VFS globals, add PTC instance injection and StateBackend read-your-writes

colifran:
### Summary
- Removed built-in readFile/writeFile globals from the QuickJS REPL. File I/O is now exclusively through PTC tools, keeping the REPL a pure JS sandbox with no implicit side effects
- ptc now accepts StructuredToolInterface instances directly alongside tool name strings. Custom tools can be injected into the REPL without being registered on the agent
- StateBackend now provides read-your-writes semantics within a single js_eval superstep by reading via __pregel_read with fresh=true. Pregel already tracks pending sends in task.writes; asking for a fresh read applies them through the reducer, removing the need for a manual write buffer
- Removed ptc: boolean shorthand from QuickJSMiddlewareOptions; use the array or object forms instead

### Tests
  - Added 12 unit tests to state.test.ts covering write-then-read, chained edits, ls/grep/glob/readRaw visibility, upload/download, duplicate write error, edit-nonexistent, and superstep boundary behavior (commitSends/clearPending)
- Expanded middleware.test.ts with PTC instance injection coverage
- Verified end-to-end with createDeepAgent + createQuickJSMiddleware across 7 scenarios (write-read, write-edit-read, write-ls, write-grep, write-glob, cross-eval, upload-download); traces confirmed in LangSmith


libs/deepagents/src/backends/state.ts
```diff
@@ -95,13 +97,51 @@ export class StateBackend implements BackendProtocolV2 {
...
+  /**
+   * Clear pending writes when the underlying state snapshot changes.
+   *
+   * Within a single LangGraph superstep, {@link getCurrentTaskInput}
+   * returns the same object reference. When a new superstep begins,
+   * the reference changes — pending writes have been committed by the
+   * reducer and can be discarded.
+   */
+  private invalidateOnNewSnapshot(stateRef: unknown): void {
+    if (this.lastStateRef !== undefined && stateRef !== this.lastStateRef) {
```
hntrl: There might be a better heuristic you can use instead of storing state entirely. Might be cursed but runtime.configurable.__pregel_task_id is distinct across supersteps?

hntrl: also `stateRef !== this.lastStateRef` seems brittle since we can't guarantee that object mutations aren't going to happen in runtime state between when its minted by StateGraph vs. here

hntrl (dismissed):

libs/deepagents/src/backends/state.ts
colifran: This was removed with the update made to use the pregel read channel.

libs/deepagents/src/backends/state.ts
```diff
@@ -89,17 +90,24 @@ export class StateBackend implements BackendProtocolV2 {
...
    */
-  private getFiles(): Record<string, FileData> {
+  private get files(): Record<string, FileData> {
     if (this.runtime) {
-      const state = this.runtime.state as { files?: Record<string, FileData> };
-      return state.files || {};
+      return (
+        (this.runtime.state as { files?: Record<string, FileData> }).files ?? {}
+      );
     }
```
hntrl, line 101: this.runtime.state is always going to be populated right? E.g. we're not going to hit the path that looks at pregel read?

libs/providers/quickjs/src/types.ts
```diff
@@ -1,30 +1,24 @@
...
    * - `{ exclude: string[] }` — expose all agent tools except these
    *
-   * @default false
+   * Omit to disable PTC entirely (default).
    */
-  ptc?: boolean | string[] | { include: string[] } | { exclude: string[] };
+  ptc?:
+    | (string | StructuredToolInterface)[]
+    | { include: (string | StructuredToolInterface)[] }
+    | { exclude: string[] };
```
hntrl: I ended up just removing these options in python too

libs/providers/quickjs/src/session.ts
```diff
@@ -406,7 +294,10 @@ export class ReplSession {
             try {
               const rawInput =
                 typeof input === "object" && input !== null ? input : {};
-              const result = await t.invoke(rawInput);
+              const result = await t.invoke(
+                rawInput,
+                this._evalConfig ?? undefined,
+              );
```
hntrl: this evalConfig behavior seems kind of brittle? Also am curious why its needed -- async context behavior should be populating this separately

libs/deepagents/src/backends/state.test.ts
```diff
@@ -796,4 +795,159 @@ describe("StateBackend", () => {
       expect(result.filesUpdate).toBeUndefined();
     });
   });
+
+  describe("pending writes (read-your-writes)", () => {
```
hntrl: are these tests meaningful at all if all we're doing is mocking pregel behavior? I would think we need to put this in the context of an actual graph instance

libs/deepagents/src/backends/state.ts
colifran: The branching handles the deprecated backend factory pattern. For zero args backends (new StateBackend()) we'll always hit the path that looks at pregel read.

libs/providers/quickjs/src/types.ts
colifran: Okay will remove here too.

colifran: Updated.

libs/providers/quickjs/src/session.ts
colifran: Hmm...I think this may have just been an artifact of something that wasn't working at one point and I thought it was needed. I removed this and everything worked. Good catch.

libs/deepagents/src/backends/state.test.ts
colifran: Removed and added integ tests instead.

libs/deepagents/src/backends/state.ts
hntrl: ah confusing but I understand

libs/deepagents/src/backends/state.int.test.ts
```diff
@@ -0,0 +1,85 @@
...
+const GraphState = Annotation.Root({
+  files: Annotation<Record<string, FileData>>({
+    reducer: (a, b) => ({ ...a, ...b }),
+    default: () => ({}),
+  }),
+  result: Annotation<string>({
+    reducer: (_, b) => b,
+    default: () => "",
+  }),
+});
```
hntrl: uber uber uber nit: use StateSchema (we need to deprecate annotaiton)

```suggestion
const GraphState = new StateSchema({
  files: new ReducedValue(
    z.record<string, FileData>,
    {
      reducer: (a, b) => ({ ...a, ...b }),
    }
  ),
  result: z.string(),
});
```

libs/providers/quickjs/src/middleware.ts
```diff
@@ -156,22 +110,48 @@ export async function generatePtcPrompt(
...
 }
 
+/**
+ * Returns true if the item is a StructuredToolInterface instance rather than a
+ * tool name string.
+ */
+export function isToolInstance(
+  item: string | StructuredToolInterface,
+): item is StructuredToolInterface {
+  return typeof item === "object" && item !== null && "invoke" in item;
```
hntrl: this can technically be any Runnable class -- there should be an isStructuredTool utility somewhere in core I'm pretty sure

colifran: Yeah I found it. I actually think we can just narrow by checking for `typeof string` though. Probably simpler.

.changeset/fluffy-donkeys-wear.md
```diff
@@ -0,0 +1,6 @@
+---
+"@langchain/quickjs": patch
```
hntrl: think we gotta treat this as breaking

```suggestion
"@langchain/quickjs": minor
```

hntrl (dismissed):
nice!

PR #492, opened 2026-04-27
feat(deepagents): implement file system permissions for fs middleware tools

colifran:
### Summary

- Adds a FilesystemPermission class and a permissions parameter to createDeepAgent and SubAgent that controls which filesystem operations each agent is allowed to perform
- Rules are evaluated first-match-wins with a permissive default; mode: "deny" blocks the operation before any backend call
- Permissions are closed over at tool-creation time — no runtime config threading, no ToolPolicy in langchain-core
- ls, glob, and grep enforce permissions in two passes: the base path check throws early; results are post-filtered to strip entries the caller can't read
- Subagents receive the parent agent's permissions by default; setting permissions on a SubAgent is a full replacement, not a merge
- FilesystemPermission, FilesystemPermissionOptions, FilesystemOperation, and PermissionMode are exported from the top-level package entry point

### Tests

- permissions/types.test.ts — FilesystemPermission construction, validation (rejects relative paths, .., ~), defaults
- permissions/enforce.test.ts — validatePath, globMatch (wildcards, brace expansion, dotfiles), decidePathAccess (first-match-wins, operation mismatch, multiple ops/paths)
- middleware/fs.permissions.test.ts — each fs tool (read, write, edit, ls, glob, grep) with allow/deny rules and mock backends; verifies backend is never called on a denied path; verifies execute is unaffected
- permissions/agent.permissions.test.ts — CreateDeepAgentParams.permissions type and default; wiring to createFilesystemMiddleware
- middleware/subagents.permissions.test.ts — SubAgent.permissions type; ?? resolution (inherit vs override vs own deny); createFilesystemMiddleware behavior for each resolved-permissions case


libs/deepagents/src/permissions/types.ts
```diff
@@ -0,0 +1,93 @@
...
+ *   }),
+ *   new FilesystemPermission({
+ *     operations: ["read"],
+ *     paths: ["/**"],
+ *     mode: "deny",
+ *   }),
+ * ];
+ * ```
+ */
+export class FilesystemPermission {
```
hntrl: not sure if we want to have classes just for configs (it makes more sense in python)

wdyt of just making this an interface? w/ some kind of discriminator like "type": "filesystem" or something like that

libs/deepagents/src/middleware/fs.ts
```diff
@@ -989,25 +1102,49 @@ export function createFilesystemMiddleware(
...
+      },
+      permissions,
+    ),
+    read_file: createReadFileTool(
+      backend,
+      {
+        customDescription: customToolDescriptions?.read_file,
+        toolTokenLimitBeforeEvict,
+      },
+      permissions,
```
hntrl: uber nit: we shouldn't use positional args, instead we should just collapse this into options.

if you need reason as to why, look at [any callback manager ever](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-core/src/callbacks/manager.ts)

libs/deepagents/src/permissions/types.ts
colifran: It has logic baked into the constructor though. There is additional business logic needed for validation.

libs/deepagents/src/middleware/fs.ts
colifran: Ohh yeah duh. I agree with you - I'll update

libs/deepagents/src/middleware/fs.ts
```diff
@@ -362,6 +367,67 @@ const FilesystemStateSchema = new StateSchema({
...
+  if (decidePathAccess(rules, operation, canonical) === "deny") {
+    throw new Error(
+      `Error: permission denied for ${operation} on ${canonical}`,
+    );
+  }
+}
+
+/**
+ * Filter a list of filesystem entries to those the rules permit.
+ *
```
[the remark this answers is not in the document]
colifran: Updated.

libs/deepagents/src/middleware/fs.ts
colifran: Fixed.

libs/deepagents/src/permissions/types.ts
colifran: Okay cleaned this up - you were right this is much better.

libs/deepagents/src/middleware/fs.ts
```diff
@@ -979,8 +1130,28 @@ export function createFilesystemMiddleware(
...
+    validatePermissionPaths(permissions);
+  }
+
+  if (
+    permissions.length > 0 &&
+    typeof backend !== "function" &&
+    isSandboxBackend(backend) &&
+    !allPathsScopedToRoutes(permissions, backend)
+  ) {
+    throw new Error(
```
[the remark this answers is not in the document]
colifran: I think this is a valid vulnerability. Unfortunately, we can't resolve the backend at construction time when it is a factory function. The fix I implemented was to catch it inside of the execute tool logic. This should successfully prevent execute from being used with permissions for backend factories.

libs/deepagents/src/middleware/fs.ts
```diff
@@ -965,6 +1076,46 @@ export interface FilesystemMiddlewareOptions {
...
+
+/**
+ * Returns true only when backend exposes route prefixes (CompositeBackend) and
+ * every permission path is scoped under one of them.
+ */
+function allPathsScopedToRoutes(
+  permissions: FilesystemPermission[],
+  backend: AnyBackendProtocol,
+): boolean {
+  if (!CompositeBackend.isCompositeBackend(backend)) {
```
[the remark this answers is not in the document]
colifran: Updated - normalized the prefix to always end with / before comparing and added a unit test to validate.

libs/deepagents/src/middleware/fs.ts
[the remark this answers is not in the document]
colifran: True positive, confirmed resolved.

Factory backends can't be resolved at construction time, so the startup guard correctly skips them. The fix adds an equivalent check inside the execute tool handler that fires after the factory is resolved at invocation time. This is the same semantics, just deferred. The tool returns a clear error and does not proceed to execution. A test covers the factory sandbox + permissions case explicitly.

The follow-up suggestions (caching, race-safety, centralization) don't apply here. Execute is the only execution-capable tool, the check is idempotent so caching adds no value, and the guard fires before any execution so concurrent calls are safe by construction.

libs/deepagents/src/backends/composite.ts
```diff
@@ -65,6 +65,23 @@ export class CompositeBackend implements BackendProtocolV2 {
...
+   * Type guard — returns true if `backend` is a {@link CompositeBackend}.
+   *
+   * Uses duck-typing on `routePrefixes` so it works across module boundaries
+   * where `instanceof` may fail.
+   */
+  static isCompositeBackend(backend: unknown): backend is CompositeBackend {
+    return Array.isArray(
+      (backend as unknown as CompositeBackend).routePrefixes,
+    );
+  }
```
hntrl: technically unsafe? Also common convention is `isInstance`

```suggestion
  static isInstance(backend: unknown): backend is CompositeBackend {
    return typeof backend === "object" && object !== null && Array.isArray(backend.routePrefixes)
    return Array.isArray(
      (backend as unknown as CompositeBackend).routePrefixes,
    );
  }
```

libs/deepagents/src/middleware/fs.ts
```diff
@@ -925,6 +1039,20 @@ function createExecuteTool(
...
+      // at configuration time — this catches that case at invocation.
+      if (
+        permissions.length > 0 &&
+        !allPathsScopedToRoutes(permissions, resolvedBackend)
+      ) {
+        return (
+          "Error: Execution not available. Filesystem permissions cannot be " +
+          "used with a backend that supports command execution because shell " +
+          "commands can access any path, making path-based rules ineffective."
+        );
```
hntrl, line 1053: we might return a ToolMessage with status="error" instead of this. Some models treat that status flag differently

libs/deepagents/src/permissions/agent.permissions.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    const params: Partial<CreateDeepAgentParams> = { name: "agent" };
+    expect(params.permissions).toBeUndefined();
+  });
+
+  it("defaults to no restrictions when omitted (empty array = allow all)", () => {
+    // Mirrors the default in createDeepAgent: permissions = []
+    const params: Partial<CreateDeepAgentParams> = {};
+    const effectivePermissions = params.permissions ?? [];
+    expect(effectivePermissions).toEqual([]);
+  });
```
hntrl: are these just meant as type tests? If so [vitest type tests](https://vitest.dev/guide/testing-types.html) may be of interest

libs/providers/quickjs/src/middleware.ts
```diff
@@ -211,14 +213,15 @@ export function createQuickJSMiddleware(
   const jsEvalTool = tool(
     async (input, config: LangGraphRunnableConfig) => {
       const threadId = config.configurable?.thread_id || DEFAULT_SESSION_ID;
+      const sessionKey = `${threadId}:${middlewareId}`;
```
hntrl, line 216: what prompted this

libs/deepagents/src/middleware/fs.ts
colifran: We do this same thing throughout the rest of the fs middleware. Why would we do something different here?

libs/providers/quickjs/src/middleware.ts
colifran: Subagents and the need for permission overrides at the subagent level. A subagent can run under the same thread_id as the parent agent but with its own createQuickJSMiddleware instance but with potentially with different permissions.

Without middlewareId in the key both would resolve to the same session and the subagent would inherit the parent's REPL context which is wrong. The middlewareId scopes each session to its specific middleware instance so parent and subagent get isolated REPLs even within the same thread.

libs/deepagents/src/permissions/agent.permissions.test.ts
colifran: The compiler already enforces that permissions accepts the right shape just going to remove them.

libs/deepagents/src/backends/composite.ts
colifran: Updated.

libs/deepagents/src/backends/composite.ts
```diff
@@ -65,6 +65,24 @@ export class CompositeBackend implements BackendProtocolV2 {
...
+  /** Route prefixes registered on this backend (e.g. `["/workspace"]`). */
+  get routePrefixes(): string[] {
+    return Object.keys(this.routes);
+  }
+
+  /**
+   * Type guard — returns true if `backend` is a {@link CompositeBackend}.
+   *
+   * Uses duck-typing on `routePrefixes` so it works across module boundaries
+   * where `instanceof` may fail.
```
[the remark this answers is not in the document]
colifran: Yes this is valid. This has been updated.

libs/deepagents/src/middleware/fs.ts
hntrl: hah ok probably fine then. We should fix this though

hntrl (approved):

PR #493, opened 2026-04-28
feat(deepagents): add handleToolErrors option to recover from tool exceptions

### LINKED ISSUE #485: feat: handle_tool_errors — JS equivalent of Python's ToolNode(handle_tool_errors=True)
## Problem

Python LangGraph has `ToolNode(handle_tool_errors=True)` which catches exceptions inside tool functions and returns them as `ToolMessage` content instead of crashing the superstep. The JS/TypeScript `deepagentsjs` package has no equivalent.

When any MCP tool throws (timeout, connection error, validation error), LangGraph JS propagates it as a **"Multiple errors occurred during superstep N"** crash. The agent cannot recover because it never sees the error in a ToolMessage — the graph just dies.

Real example from production:

```
Error: MCP tool 'wam_load_chain' on server 'wam' returned an error: 
Timeout: browser did not respond to "loadPlugin" within 10s
→ Multiple errors occurred during superstep 9.
```

## Current workaround

Every tool must manually wrap its body in try/catch and return `{ ok: false, error, hint }` as JSON text content instead of throwing. This is boilerplate that belongs in the framework:

```ts
// What every tool currently has to do manually
async (params) => {
  try {
    const result = await callBrowser(...)
    return { content: [{ type: "text", text: JSON.stringify(result) }] }
  } catch (err) {
    return { content: [{ type: "text", text: JSON.stringify({ ok: false, error: err.message, hint: "retry..." }) }] }
  }
}
```

## Requested API

```ts
// Option A — createDeepAgent config
createDeepAgent({ tools, handleToolErrors: true })

// Option B — ToolNode level (matches Python API)
new ToolNode(tools, { handleToolErrors: true })
```

When enabled: any exception thrown inside a tool function is caught, serialized as a `ToolMessage` with content `"Error: <message>"`, and the graph continues — giving the LLM a chance to read the error and retry with a corrected call.

## Related

- deepagents #947 — validation errors not retried (model responds with plain text instead of corrected tool call)
- deepagentsjs #68 — original feature request, no resolution
- langgraph #6486 — `handle_tool_errors` disabled by default after 1.0.1, breaking existing graphs
- [LangGraph forum: self-healing agents via tool error messages](https://forum.langchain.com/t/raising-tool-call-errors-so-agents-can-be-self-healing/3152)

JadenKim-dev:
## Problem

When an tool throws (timeout, connection error, validation error),
LangGraph JS propagates it as a "Multiple errors occurred during superstep N"
crash. The agent never sees the error in a ToolMessage and cannot recover.

## Solution

Adds opt-in `handleToolErrors` to `createDeepAgent`. When enabled, exceptions
thrown inside tool functions are caught and returned as a `ToolMessage` with
`status: "error"` instead of crashing the graph.

## Usage

​```ts
  tools: [myTool],
  handleToolErrors: true,
})
​```

The LLM sees the error message and can decide whether to retry with different
arguments.

## Implementation

A new `createToolErrorHandlerMiddleware()` is inserted as the outermost
`wrapToolCall` layer in the middleware stack, so it catches any exception
thrown anywhere in the tool execution chain. `NodeInterrupt` (HITL breakpoints)
and aborted `AbortSignal` are always re-thrown. Defaults to `false` to preserve
existing behavior.

Closes #485

### DIFF
--- a/libs/deepagents/src/agent.test.ts
+++ b/libs/deepagents/src/agent.test.ts
@@ -1,12 +1,19 @@
 import { describe, it, expect, vi } from "vitest";
 import { createDeepAgent, isAnthropicModel } from "./agent.js";
-import { FakeListChatModel } from "@langchain/core/utils/testing";
 import {
+  FakeListChatModel,
+  FakeStreamingChatModel,
+} from "@langchain/core/utils/testing";
+import {
+  AIMessage,
   HumanMessage,
   SystemMessage,
+  ToolMessage,
   type BaseMessage,
 } from "@langchain/core/messages";
 import { MemorySaver } from "@langchain/langgraph";
+import { tool } from "langchain";
+import { z } from "zod";
 import { createFileData } from "./backends/utils.js";
 import { ConfigurationError } from "./errors.js";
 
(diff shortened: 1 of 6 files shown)


hntrl (commented):
Thanks for the contribution! Going to close because we already have guidance on how to handle tool errors using middleware. See here: https://docs.langchain.com/oss/javascript/langchain/tools#error-handling

PR #497, opened 2026-04-30
feat(quickjs): add maxPtcCallsbudget for ptc calls

colifran:
Port of https://github.com/langchain-ai/deepagents/pull/2994

### DIFF
--- a/libs/providers/quickjs/src/session.ts
+++ b/libs/providers/quickjs/src/session.ts
@@ -35,6 +35,7 @@ export const DEFAULT_MEMORY_LIMIT = 50 * 1024 * 1024;
 export const DEFAULT_MAX_STACK_SIZE = 320 * 1024;
 export const DEFAULT_EXECUTION_TIMEOUT = 30_000;
 export const DEFAULT_SESSION_ID = "__default__";
+export const DEFAULT_MAX_PTC_CALLS = 256;
 
 // The variant descriptor (WASM binary + glue) is safe to share across sessions;
 // only the instantiated module carries asyncify state. Import once, instantiate per session.
--- a/libs/providers/quickjs/src/session.ts
+++ b/libs/providers/quickjs/src/session.ts
@@ -352,8 +352,6 @@ export class ReplSession {
     const err = new Error(
       `PTC call budget of ${this.maxPtcCalls} exceeded — attempted call: tools.${functionName}`,
     );
-    err.name = "PTCCallBudgetExceeded";
-
     throw err;
   }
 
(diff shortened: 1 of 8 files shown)


libs/providers/quickjs/src/session.ts
```diff
@@ -317,6 +324,39 @@ export class ReplSession {
...
+    if (this.ptcCallsRemaining > 0) {
+      this.ptcCallsRemaining--;
+      return;
+    }
+
+    const err = new Error(
+      `PTC call budget of ${this.maxPtcCalls} exceeded — attempted call: tools.${functionName}`,
+    );
+    err.name = "PTCCallBudgetExceeded";
+
```
hntrl: can we hoist this out into its own error class? src/errors.ts or something like that

libs/providers/quickjs/src/session.ts
```diff
@@ -395,70 +435,75 @@ export class ReplSession {
...
+      }
+
+      result.value.dispose();
+      return {
+        ok: false,
+        error: { message: "Promise timed out — execution interrupted" },
+        logs: [...this.logs],
+      };
+    } finally {
+      this.ptcCallsRemaining = null;
```
hntrl, line 503: do you want to reset here instead of setting to null?

libs/providers/quickjs/src/session.ts
colifran: Should we? We don't normally have specific error classes. I meant to remove err.name - I was just throwing this and letting the existing catch logic handle it.

libs/providers/quickjs/src/session.ts
colifran: No, the reset happens at the start of the eval. This is the same concept used in the Python version: https://github.com/langchain-ai/deepagents/pull/2994/changes#diff-3a46ed05953b09ef925c369613036a9c7d9081a7a1edcf2fd061efa6d3e2f975R649

libs/providers/quickjs/src/session.ts
colifran: Ends up looking like:
```
FAILED: {"message":"Tool 'greet' failed: PTC call budget of 2 exceeded — attempted call: tools.greet","name":"Error","stack":"    at greet (native)\n    at <anonymous> (eval.js:6:9)\n    at <eval> (eval.js:18:1)\n"}
```

hntrl: It's probably something we should do more often

hntrl (dismissed):
approved

libs/providers/quickjs/src/session.ts
colifran: Updated!

PR #501, opened 2026-05-01
fix: bump @langchain/core to ^1.1.42 across all workspace packages

sukhmanghotraa:
## Problem

When using `@langchain/google` as a model provider in the examples workspace, the server crashes immediately with:

1. Runtime crash on startup:

    ```bash
    ERR_PACKAGE_PATH_NOT_EXPORTED: Package subpath './utils/uuid' is not
    defined by "exports" in @langchain/core/package.json
    ```

    `@langchain/google` imports `@langchain/core/utils/uuid`, a subpath added in `@langchain/core@1.1.42`. Both the examples workspace and `libs/deepagents` were pinned to `^1.1.40` which does not export this path.

 2. TypeScript type error after fixing the runtime crash:
 
    ```bash
    Type 'ChatGoogle' is not assignable to type 'BaseLanguageModel<any, BaseLanguageModelCallOptions>'
    ```
    Even after bumping `@langchain/core` in `examples/` only, pnpm kept two separate copies in its store - `1.1.40` for `deepagents` and `1.1.42` for `@langchain/google`. TypeScript treats them as structurally incompatible types because the type import paths differ.

## Fix

- Bump `@langchain/core` from `^1.1.40` → `^1.1.42` in all `../package.json` files

## Testing

```bash
cd examples/research
pnpm dlx @langchain/langgraph-cli dev
# -> No ERR_PACKAGE_PATH_NOT_EXPORTED and agent starts correctly

### DIFF
--- a/.changeset/brown-rings-sparkle.md
+++ b/.changeset/brown-rings-sparkle.md
@@ -1,0 +1,7 @@
+---
+"deepagents-acp": patch
+"deepagents": patch
+"@langchain/quickjs": patch
+---
+
+fix: bump @langchain/core to ^1.1.42 across all workspace packages
--- a/examples/async-subagents/parallel-research/package.json
+++ b/examples/async-subagents/parallel-research/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "deepagents": "workspace:*",
     "@langchain/langgraph": "^1.2.9",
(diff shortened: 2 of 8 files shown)


hntrl (approved):
thanks for the fix 🙏🙏

PR #541, opened 2026-05-17
fix(deepagents): return application/octet-stream for unknown MIME types

### LINKED ISSUE #542: getMimeType returns text/plain for binary files (.zip, .exe, .wasm, .sqlite, etc.)
getMimeType() in backends/utils.ts defaults to "text/plain" for any extension not in the MIME_TYPES map. The map only covers images, audio, video, and PDFs — no text/code extensions either, everything just falls through to the default.

The problem is isTextMimeType("text/plain") returns true, so binary files get treated as text across the board:

- grep searches through .zip, .exe, .wasm, .sqlite, .pyc, .class, .jar files and returns garbage matches
- read_file splits binary content on \n and paginates it as "lines"
- StateBackend.write() passes binary data through createFileData() as text
- StateBackend.glob() reports file sizes from .content.length on garbled strings

```ts
// backends/utils.ts line 748-751
export function getMimeType(filePath: string): string {
  const ext = path.extname(filePath).toLocaleLowerCase();
  return MIME_TYPES[ext] || "text/plain";  // <-- this default is wrong
}
```

Every caller that does `if (!isTextMimeType(mimeType)) { skip }` — like grepSearchFiles, grepMatchesFromFiles, etc. — never skips binary files because they all come back as text/plain.

The default should be "application/octet-stream". isTextMimeType already returns false for it.

ixchio:
Fixes #542.

`getMimeType()` falls back to `"text/plain"` for any extension not in the `MIME_TYPES` lookup table. Since `isTextMimeType("text/plain")` returns `true`, this means binary files (.zip, .exe, .wasm, .sqlite, .pyc, .class, .jar, etc.) get treated as text everywhere downstream — grep searches through them and returns garbage, read_file splits their content on newlines and paginates it, and StateBackend can silently corrupt binary data stored via createFileData().

The fix is straightforward: change the default to `"application/octet-stream"`, which is what unknown binary data should be. `isTextMimeType()` already returns `false` for it, so binary files stop getting mangled.

To avoid breaking existing behavior for actual text files that were previously relying on the fallback (like .txt, .ts, .py, .json, .js — none of which had explicit entries), this also adds those to the MIME_TYPES map directly. Covers ~67 common text/code extensions.

All 386 backend tests pass (10 files). The updated test expectations in utils.test.ts verify both sides: known text extensions still resolve correctly, and unknown/binary extensions now get `application/octet-stream` instead of being misidentified as text.


libs/deepagents/src/backends/utils.ts
```diff
@@ -69,6 +69,73 @@ const MIME_TYPES: Record<string, string> = {
...
+  ".vue": "text/plain",
+  ".svelte": "text/plain",
+  ".astro": "text/plain",
+  ".tf": "text/plain",
+  ".cmake": "text/plain",
+  ".makefile": "text/plain",
+  ".dockerfile": "text/plain",
+  ".gitignore": "text/plain",
+  ".dockerignore": "text/plain",
+  ".editorconfig": "text/plain",
```
christian-bromann, line 137: Instead of maintaining this extensive list of mime types I think it makes sense to just embed [`mime-types`](https://www.npmjs.com/package/mime-types?activeTab=readme)?

ixchio: Looked into it turns out mime-types thinks .ts is an MPEG transport stream (video/mp2t) which would be fun to explain to every TypeScript user whose files suddenly get treated as binary ans Also .jsx, .vue, .svelte, .astro straight up don't exist in IANA, so they'd fall to octet-stream. 
Happy to switch to a hybrid approach though mime-types as the primary lookup with a small override map for the ones it gets wrong. 

Let me know what you'd prefer.

hntrl (dismissed):

hntrl (approved):
thanks for the pr!

PR #551, opened 2026-05-22
fix(deepagents): gate cache_control writes on per-call request.model

### LINKED ISSUE #550: cache_control leaks to non-Anthropic providers on modelFallbackMiddleware swap
## Summary

`createCacheBreakpointMiddleware` and `createMemoryMiddleware` write Anthropic-specific `cache_control: { type: "ephemeral" }` markers into the request payload based on a **boot-time** decision in `createDeepAgent` ("is the primary model Anthropic?"). When LangChain's `modelFallbackMiddleware` swaps `request.model` to a non-Anthropic provider mid-flight (e.g. on Anthropic 5xx), the markers leak through and the fallback provider rejects the request:

```
400 Unknown parameter: 'messages[0].content[N].cache_control'
```

The fallback chain that exists *precisely* to survive Anthropic outages is destroyed by a payload-shape bug. End users see a 5xx instead of a slower-but-correct answer from the fallback model.

## Reproduction

Minimal: create an agent with Anthropic primary + OpenAI fallback via `modelFallbackMiddleware`. Force the primary call to fail. Capture the OpenAI request body — it carries `cache_control` blocks. OpenAI 400s on the unknown field.

Production trace where this surfaced: Anthropic returned 529 `overloaded_error`, `modelFallbackMiddleware` walked Opus → Sonnet → gpt-5.2, every gpt-5.2 attempt failed with the 400 above.

## Root cause

Two writers, both with boot-time gating but no per-call check:

1. **`libs/deepagents/src/middleware/cache.ts`** — `createCacheBreakpointMiddleware().wrapModelCall` tags `systemMessage.content[N]` with `cache_control` on every call. No `request.model` inspection. Installed conditionally in `agent.ts:218-226` when `isAnthropicModel(model)` is true at boot.

2. **`libs/deepagents/src/middleware/memory.ts:342-344`** — writes `cache_control` onto the memory block, gated on the `addCacheControl` flag set once at agent creation (`agent.ts:375`: `addCacheControl: anthropicModel`).

Both gates are correct at boot. Neither survives `modelFallbackMiddleware` swapping `request.model` at request time.

## Why fix it in the writers

`modelFallbackMiddleware` (LangChain side) spreads the request shape blindly: `handler({ ...request, model })`. It can't strip provider-specific markers because it doesn't know which fields are provider-specific — that knowledge lives in the writer.

The classification helper already exists: `isAnthropicModel` in `libs/deepagents/src/utils.ts:9-18`, which handles string / `ConfigurableModel` / `ChatAnthropic` instances. Reusing it inside `wrapModelCall` is one `if` statement per writer.

## Proposed fix

\`\`\`ts
// cache.ts — inside wrapModelCall, before existing logic
if (!isAnthropicModel(request.model)) return handler(request);

// memory.ts — replace the boot-time \`addCacheControl\` flag with a
// per-call check on request.model
\`\`\`

Boot-time gate in `agent.ts` can stay (small perf win when primary is non-Anthropic — middleware isn't installed at all). The per-call check covers the fallback case where the middleware *is* installed but the model has been swapped.

## Related

- `langchain-ai/deepagentsjs#280` — different toggle, but same area
- `langchain-ai/langchain#33709` — Python, same bug class

## Workaround for users hitting this today

Strip `cache_control` from the request inside a custom middleware positioned immediately inside `modelFallbackMiddleware`. Not pretty but unblocks production until this lands.

## Out of scope for this issue

`request.modelSettings.cache_control` (written by `anthropicPromptCachingMiddleware` in LangChain core) can also leak across the swap. That's a LangChain-side concern, not deepagents.

antonnak:
Fixes #550.

## Problem

`createCacheBreakpointMiddleware` and `createMemoryMiddleware` write Anthropic-specific `cache_control: { type: "ephemeral" }` markers into the request payload. Both gate the write at **agent-creation time** in `createDeepAgent` (whether the *primary* model is Anthropic). When `modelFallbackMiddleware` swaps `request.model` to a non-Anthropic provider at request time, the markers leak through and the fallback provider rejects the request:

```
400 Unknown parameter: 'messages[0].content[N].cache_control'
```

Full root-cause writeup, production trace, and reproduction in #550.

## Fix

Move the model classification from boot-time to **per-call**, using the existing `isAnthropicModel` helper (`src/utils.ts:9`).

- **`src/middleware/cache.ts`** — bail at the top of `wrapModelCall` when `request.model` isn't Anthropic.
- **`src/middleware/memory.ts`** — AND the existing `addCacheControl` flag with the same per-call check. `addCacheControl` keeps its current public meaning (opt-in switch); the per-call gate is an additional safety net.
- **`src/agent.ts`** — left untouched. The boot-time install gate stays (small perf win when primary is non-Anthropic — middleware isn't installed at all). The new per-call check handles the fallback case where the middleware *is* installed but the model has been swapped.

## Tests

Each middleware's `.test.ts` got a `per-call provider gating` block. They build a request with a `ChatAnthropic`-shaped stub vs a `ChatOpenAI`-shaped stub on `request.model`, run `wrapModelCall`, and assert `cache_control` is/isn't written. Also covers `ConfigurableModel` with `modelProvider: "anthropic"` and the string forms (`"anthropic:claude-..."`, `"openai:gpt-5"`).

Existing fixtures that didn't set `request.model` now pass a `ChatAnthropic` stub so they keep exercising the write path — they were previously relying on the unconditional behavior. No behavioral assertions changed; only the fixture got more explicit.

31/31 tests pass on `libs/deepagents` (`cache.test.ts` + `memory.test.ts`).

## Compatibility

- **No public API change.** `addCacheControl?: boolean` on `MemoryMiddlewareOptions` keeps its existing semantics.
- **Anthropic-primary happy path unchanged** — same writes, same cache hits.
- **Per-call cost:** one `getName()` / string check inside `wrapModelCall`. Sub-microsecond.
- **Failure mode improvement:** users running `createDeepAgent` with an Anthropic primary + non-Anthropic fallback no longer get a 400 from the fallback provider on swap.

## Changeset

`patch` bump on `deepagents`.

## Out of scope

`request.modelSettings.cache_control` (written by LangChain core's `anthropicPromptCachingMiddleware`) can also leak across the swap. Mentioned in #550 for tracking — that's a LangChain-side fix, not deepagents.

### DIFF
--- a/.changeset/cache-control-fallback-leak.md
+++ b/.changeset/cache-control-fallback-leak.md
@@ -1,0 +1,13 @@
+---
+"deepagents": patch
+---
+
+fix(deepagents): gate cache_control writes on per-call request.model
+
+`createCacheBreakpointMiddleware` and `createMemoryMiddleware` were gating
+the Anthropic-specific `cache_control` write at agent-creation time only.
+When `modelFallbackMiddleware` swapped `request.model` to a non-Anthropic
+provider mid-flight (e.g. on Anthropic 5xx), the marker leaked through
+and the fallback provider rejected the request with
+`400 Unknown parameter: 'cache_control'`. Both middlewares now also
+check `isAnthropicModel(request.model)` inside `wrapModelCall`. Fixes #550.
--- a/libs/deepagents/src/middleware/cache.test.ts
+++ b/libs/deepagents/src/middleware/cache.test.ts
@@ -3,13 +3,22 @@ import { createCacheBreakpointMiddleware } from "./cache.js";
 import { SystemMessage } from "@langchain/core/messages";
 
 describe("createCacheBreakpointMiddleware", () => {
+  // Real wrapModelCall calls always receive a model. The middleware now
+  // gates per-call on `isAnthropicModel(request.model)` (#550), so every
+  // existing test gets a ChatAnthropic stub to keep the cache_control
+  // writes on the path under test.
+  const fakeAnthropicModel = { getName: () => "ChatAnthropic" };
+
   describe("wrapModelCall", () => {
     it("adds cache_control to the last block of a string system message", () => {
       const middleware = createCacheBreakpointMiddleware();
       const mockHandler = vi.fn().mockReturnValue({ response: "ok" });
 
       middleware.wrapModelCall!(
-        { systemMessage: new SystemMessage("Base prompt") } as any,
+        {
+          model: fakeAnthropicModel,
+          systemMessage: new SystemMessage("Base prompt"),
+        } as any,
         mockHandler,
       );
 
@@ -25,6 +34,7 @@ describe("createCacheBreakpointMiddleware", () => {
 
       middleware.wrapModelCall!(
         {
+          model: fakeAnthropicModel,
           systemMessage: new SystemMessage({
             content: [
               { type: "text", text: "Block 1" },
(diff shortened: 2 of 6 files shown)


hntrl (approved):
thanks for the pr!

PR #554, opened 2026-05-23
fix(quickjs): forward config for skills and PTC tools

### LINKED ISSUE #553: QuickJS middleware drops LangGraph config for skillsBackend and PTC tool calls
## Summary
`createCodeInterpreterMiddleware()` has two paths where the LangGraph tool `config` is available but not forwarded:
1. `skillsBackend` setup calls `prepareSkillsForEval()` without config, causing `getCurrentTaskInput()` to fall back to `AsyncLocalStorage`.
2. Programmatic Tool Calling inside the QuickJS REPL invokes bridged tools with `t.invoke(rawInput)` instead of `t.invoke(rawInput, config)`.
This breaks browser/web builds, or any environment without working `AsyncLocalStorage`, and also breaks bridged tools that need graph runtime context, such as DeepAgents’ `task` tool.
## Versions
- `@langchain/quickjs`: 0.4.0
- `deepagents`: 1.10.2
- `@langchain/langgraph`: 1.3.2
## Environment
Browser / webpack build.
## Reproduction
```ts
import { createDeepAgent, StateBackend } from "deepagents";
import { createCodeInterpreterMiddleware } from "@langchain/quickjs";
// Created this way because there is a bug when
// we use only `new StateBackend()`, see https://github.com/langchain-ai/deepagentsjs/issues/547
const stateBackend = (runtime) => new StateBackend(runtime);
  model: "openai:gpt-5.4",
  backend: stateBackend,
  skills: ["/skills/"],
  middleware: [
    createCodeInterpreterMiddleware({
      skillsBackend: stateBackend,
      ptc: ["task"],
    }),
  ],
});
```

Then trigger either:
- the interpreter tool with skills enabled
- a PTC call from inside the interpreter, for example:

```js
await tools.task({
  description: "...",
  subagent_type: "seo-agent",
});
```

## Actual behavior
The skillsBackend path calls:

`const taskInput = getCurrentTaskInput();`

and the PTC bridge calls:

`const result = await t.invoke(rawInput);`

In the first case, `getCurrentTaskInput()` falls back to `AsyncLocalStorage`:

`const runConfig = config ?? AsyncLocalStorageProviderSingleton.getRunnableConfig();`

In browser/web environments this can throw:
```
Config not retrievable. This is likely because you are running in an environment without support for AsyncLocalStorage.
If you're running `getCurrentTaskInput` in such environment, pass the `config` from the node function directly.
```

In the PTC case, tools invoked through tools.* do not receive graph runtime config. For example, await tools.task(...) can fail with:

`Error: Tool 'task' failed: BUG: internal scratchpad not initialized.`

## Expected behavior
The middleware should forward the existing tool config in both places:
- `prepareSkillsForEval()` should call `getCurrentTaskInput(config)`
- PTC tools invoked from the REPL should receive config via `t.invoke(rawInput, config)`

## Root cause
Current quickjs skills flow drops config before reading task input:
```js
tool(async (input, config) => {
  if (skillsBackend !== undefined) {
    const setupError = await prepareSkillsForEval(
      session,
      skillsBackend,
      input.code
    );
  }
});
async function prepareSkillsForEval(session, skillsBackend, code) {
  const taskInput = getCurrentTaskInput();
}
```

Current PTC bridge also drops config:
`const result = await t.invoke(rawInput);`

## Proposed fix
Thread config through the skills setup:
```js
async function prepareSkillsForEval(session, skillsBackend, code, config) {
  const taskInput = getCurrentTaskInput(config);
  const resolved = await resolveBackend(skillsBackend, {
    ...config,
    state: taskInput,
  });
}
```

Store the active eval config on the `ReplSession` and forward it to bridged tools:
```js
session.setToolConfig(config);
const result = await t.invoke(rawInput, this.toolConfig);
```

## Why this matters
This is especially important for browser/web builds, where `AsyncLocalStorage` may not exist or may be stubbed out. It also ensures PTC tools called from inside QuickJS receive the same graph context as direct top-level tool calls.

luizzappa:
## Summary

Fixes #553.

`createCodeInterpreterMiddleware()` had two places where the LangGraph runnable config was available but not forwarded:

- the `skillsBackend` setup path
- Programmatic Tool Calling from inside the QuickJS REPL, e.g. `await tools.task(...)`

In the first case, `getCurrentTaskInput()` fell back to `AsyncLocalStorage`, which breaks in browser/web environments where `AsyncLocalStorage` is unavailable or stubbed.

In the second case, bridged tools were invoked without graph runtime config, which breaks tools that need current graph state, such as DeepAgents’ `task` tool.

## Changes

- Pass config from the code interpreter tool handler into `prepareSkillsForEval()`
- Use `getCurrentTaskInput(config)` in the quickjs skills backend path
- Store the active eval tool config on `ReplSession`
- Forward that config to PTC tool invocations via `t.invoke(rawInput, config)`
- Add coverage proving a PTC tool invoked through `tools.*` receives the eval config

## Why

This allows QuickJS skills and PTC tools to work in browser/web builds and preserves graph runtime context for tools called from inside the interpreter.

### DIFF
--- a/libs/providers/quickjs/src/session.ts
+++ b/libs/providers/quickjs/src/session.ts
@@ -26,6 +26,7 @@ import type {
   QuickJSAsyncWASMModule,
 } from "quickjs-emscripten-core";
 import type { StructuredToolInterface } from "@langchain/core/tools";
+import type { RunnableConfig } from "@langchain/core/runnables";
 
 import { loadSkill, scanSkillReferences, type LoadedSkill } from "./skills.js";
 import { PTCCallBudgetExceededError } from "./errors.js";
@@ -309,6 +310,7 @@ export class ReplSession {
   private skillsFailed: Map<string, Error> = new Map();
   private readonly maxPtcCalls: number | null;
   private ptcCallsRemaining: number | null = null;
+  private toolConfig: RunnableConfig | undefined;
 
   /**
    * Reset the shared WASM module. Forces the next session to instantiate
@@ -636,6 +638,14 @@ export class ReplSession {
   }
 
   /**
+   * Store the active LangGraph tool config for the current eval invocation.
+   * Programmatic tool calls inside QuickJS forward this config to bridged tools.
+   */
+  setToolConfig(config?: RunnableConfig): void {
+    this.toolConfig = config;
+  }
+
+  /**
    * Evaluate code in this session.
    *
    * Lazily starts the QuickJS runtime on the first call. Code is
@@ -730,6 +740,7 @@ export class ReplSession {
       };
     } finally {
       this.ptcCallsRemaining = null;
+      this.toolConfig = undefined;
     }
   }
 
@@ -813,7 +824,7 @@ export class ReplSession {
               this.consumePtcBudget(camelName);
               const rawInput =
                 typeof input === "object" && input !== null ? input : {};
-              const result = await t.invoke(rawInput);
+              const result = await t.invoke(rawInput, this.toolConfig);
               let text = extractToolText(result);
               if (t.name === "read_file") {
                 text = stripLineNumbers(text);
--- a/.changeset/rare-buses-switch.md
+++ b/.changeset/rare-buses-switch.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+pass config when preparing task tool
--- a/.changeset/strict-ads-smell.md
+++ b/.changeset/strict-ads-smell.md
@@ -1,0 +1,5 @@
+---
+"@langchain/quickjs": patch
+---
+
+Fix browser/web usage of skills backend by passing runnable config into skill preparation.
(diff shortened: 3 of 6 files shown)


libs/providers/quickjs/src/session.ts
```diff
@@ -635,6 +637,14 @@ export class ReplSession {
     this.skillsContext = ctx;
   }
 
+  /**
+   * Store the active LangGraph tool config for the current eval invocation.
+   * Programmatic tool calls inside QuickJS forward this config to bridged tools.
+   */
+  setToolConfig(config?: RunnableConfig): void {
+    this.toolConfig = config;
+  }
```
hntrl, line 646: FWIW I think this might cause some issues with how we handle parallel eval calling. Its something we have to guard against at the emscripten boundary but if we make enhancements there this becomes problematic. We'll probably be addressing sometime soon so I'll try to keep this top of mind

hntrl (approved):

bleafman:
Hello! Any blockers to merging this? I just hit this gap when trying to add tools to PTC that have different behavior based on the runtime context.

Issue #534, opened 2026-05-12
feat(streaming): v3 streamEvents causes INVALID_TOOL_RESULTS 400 and leaks tool results as text tokens


anouar-bm:
## Environment

- `deepagents`: 1.10.1
- `@langchain/langgraph`: latest
- Model: OpenAI gpt-5.4 / gpt-4o
- Node.js 20, Next.js 16 App Router

## Description

When using `streamEvents` with `version: "v3"` (introduced in #458 / `b1e1b7b`), two bugs occur on every fresh thread:

### Bug 1: `INVALID_TOOL_RESULTS` 400 error

After the first tool call completes and its `ToolMessage` is stored, the agent throws:

```
Error: 400 Invalid parameter: messages with role 'tool' must be a response to a preceding message with 'tool_calls'.
```

This does **not** happen with `version: "v2"` on the same thread and same input. The error occurs on a **fresh thread** (new `thread_id`), so it is not caused by stale checkpointer state.

### Bug 2: `run.messages` emits tool result content as text tokens

When iterating `run.messages` concurrently with `run.toolCalls`, tool result content (e.g. `"[]"`) appears as a `text` SSE event — as if the model had emitted it as an AI message. This causes the tool result to be displayed to the user as assistant text.

Observed SSE output:
```
data: {"type":"tool_start","id":"call_xxx","name":"wam_list_plugins","input":{},"agent":"wam-coordinator"}
data: {"type":"tool_end","id":"call_xxx","output":{"lc":1,"type":"constructor","id":["langchain_core","messages","ToolMessage"],...}}
data: {"type":"text","text":"[]"}   ← tool result leaked as text token
data: {"type":"error","text":"400 Invalid parameter: messages with role 'tool'..."}
```

Note also that `tool_end` output is still a serialized `ToolMessage` `{lc:1, type:"constructor", kwargs:{content:"..."}}` — not the raw tool return value as expected from the v3 typed API.

## Reproduction

```typescript
const run = await agent.streamEvents(
  { messages: [{ role: "user", content: "hello" }] },
  { version: "v3", configurable: { thread_id: crypto.randomUUID() }, recursionLimit: 100 }
);

await Promise.all([
  (async () => {
    for await (const msgStream of run.messages) {
      for await (const event of msgStream) {
        // Receives tool result content as text-delta events
        console.log("msg event", event);
      }
    }
  })(),
  (async () => {
    for await (const toolCall of run.toolCalls) {
[...]

PR #555, opened 2026-05-24
fix(deepagents): bump langchain to ^1.4.2 to fix Buffer.from(undefined) crash in v3 stream

anouar-bm:
## Summary

- Bumps `langchain` dependency from `^1.4.0` → `^1.4.2` in `libs/deepagents/package.json`
- Ensures the next build includes the fix from langchain-ai/langchainjs#10900 ("fix(langchain): unwrap tool message outputs in agent streams")

## Root cause

deepagents bundles langchain into `dist/index.js` via `tsdown`. The 1.10.2 bundle was built against `langchain ^1.4.0`, which predates the fix in 1.4.2. Upgrading langchain in a host project has no effect — the bundled copy inside `dist/index.js` still contains the pre-fix code.

## Symptom (follow-up to #534)

When using `v3` `streamEvents` and a tool call fails then retries (e.g. `write_code op=replace` after a compile error), the stream crashes with:

```
Error: The first argument must be of type string or an instance of Buffer,
       ArrayBuffer, or Array or an Array-like Object. Received undefined
    at Object.process (langchain/src/agents/stream.ts:327:36)
    at StreamMux.push (@langchain/langgraph/src/stream/mux.ts:287:24)
    at pump            (@langchain/langgraph/src/stream/mux.ts:499:13)
```

The crash fires because the pre-1.4.2 messages transformer calls `Buffer.from(content)` where `content` is `undefined` when a `ToolMessage` carries an array-typed `content` field (a `tool_call` block with no text). This is the same underlying bug as #534 — tool message content not unwrapped before the `Buffer.from` call.

## Fix

Bumping to `^1.4.2` ensures the fixed unwrap logic is compiled into the next published bundle.

## Test plan

- [ ] `pnpm build` in `libs/deepagents` resolves langchain ≥ 1.4.2
- [ ] v3 `streamEvents` no longer crashes when a tool call fails and retries in a multi-step run

### DIFF
--- a/libs/deepagents/package.json
+++ b/libs/deepagents/package.json
@@ -42,7 +42,7 @@
     "@langchain/langgraph": "^1.3.0",
     "@langchain/langgraph-sdk": "^1.9.1",
     "fast-glob": "^3.3.3",
-    "langchain": "^1.4.0",
+    "langchain": "^1.4.2",
     "micromatch": "^4.0.8",
     "yaml": "^2.8.2",
     "zod": "^4.3.6"
--- a/libs/deepagents/package.json
+++ b/libs/deepagents/package.json
@@ -39,7 +39,7 @@
   "homepage": "https://github.com/langchain-ai/deepagentsjs#readme",
   "dependencies": {
     "@langchain/core": "^1.1.44",
-    "@langchain/langgraph": "^1.3.0",
+    "@langchain/langgraph": "^1.3.2",
     "@langchain/langgraph-sdk": "^1.9.1",
     "fast-glob": "^3.3.3",
     "langchain": "^1.4.2",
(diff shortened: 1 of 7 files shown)


anouar-bm:
Hey @christian-bromann — you closed #534 and landed the fix in langchain 1.4.2 (langchain-ai/langchainjs#10900), thanks for that.

This PR is a follow-up: the fix exists in `langchain` 1.4.2 but deepagents bundles langchain into `dist/index.js` via `tsdown`, so upgrading langchain in a host project has no effect — the pre-fix code is still in the bundle. Bumping the dep to `^1.4.2` here and rebuilding is the missing step to actually ship the fix to users.

Could you take a look and confirm this is the right place to make the change? Happy to add a changelog entry or anything else needed.

PR #563, opened 2026-06-01
fix(deepagents): isolate subagent projections per task invocation

### LINKED ISSUE #560: `stream.subagents` shares `messages` / `toolCalls` between parallel same-type subagent invocations
## Description
When consuming Deep Agents event streams with `stream.subagents`, parallel invocations of the same subagent type appear to share the same `messages` and `toolCalls` projections.

For example, if the coordinator calls two `task` tools in parallel with the same `subagent_type`, each returned `subagent.toolCalls` stream can receive tool calls from both subagent invocations. This makes the UI show duplicated or cross-contaminated tool calls.

## Expected behavior
Each `SubagentRunStream` yielded by `stream.subagents` should expose only the messages/tool calls for that specific subagent invocation.

```ts
for await (const subagent of stream.subagents) {
  void (async () => {
    for await (const call of subagent.toolCalls) {
      // Expected: only tool calls for this specific subagent invocation
      console.log(subagent.name, call.name);
    }
  })();
}
```

## Actual behavior

When two same-type subagents run in parallel:
- subagent A’s toolCalls stream receives A and B tool calls
- subagent B’s toolCalls stream receives A and B tool calls
- same risk appears to apply to messages

## Suspected cause
In `createSubagentTransformer`, subagent stream logs appear to be keyed by `subagentName` rather than by a unique subagent invocation id.

That means parallel invocations with the same `subagent_type` reuse the same internal `StreamChannels`:
- messagesLog
- toolCallsLog
- nestedSubagentsLog

So multiple `SubagentRunStreams` can point at the same channels.

## Suggested fix
Key subagent projection state by a unique invocation key, such as the `task` tool call id, instead of by subagent name.

For child tool calls, also key pending tool-call state by both subagent invocation and child tool call id, to avoid collisions across sibling subagents.

Conceptually:
```js
const subagentKey = toolCallId ? `task:${toolCallId}` : `subagent:${nextSubagentKey++}`;
subagentsByKey.set(subagentKey, logs);
toolsNodeToKey.set(namespaceSegment, subagentKey);
const childToolCallKey = `${subagentKey}:${toolCallId}`;
```

## Environment
- Package: `deepagents`
- Version: `1.10.2`
- Runtime: JavaScript / TypeScript
- Streaming API: `agent.streamEvents(input, { version: "v3" })`

## Additional context
This is especially visible in live UIs that consume coordinator and subagent projections concurrently, as recommended in the event streaming docs.

christian-bromann:
## Summary
- Fix `createSubagentTransformer` so each `task` invocation gets its own `messages` / `toolCalls` / nested `subagents` channels, keyed by `task:${tool_call_id}` instead of `subagent_type`.
- Route child-namespace events via `toolsNodeToKey` and scope pending child tool state with `${invocationKey}:${toolCallId}` to avoid collisions across sibling subagents.
- Add a unit regression test for two parallel `researcher` invocations with isolated ping tool-call routing.

fixes #560

### DIFF
--- a/.changeset/fix-subagent-stream-invocation-keys.md
+++ b/.changeset/fix-subagent-stream-invocation-keys.md
@@ -1,0 +1,9 @@
+---
+"deepagents": patch
+---
+
+fix(stream): isolate subagent projections per task invocation
+
+Key `createSubagentTransformer` channels by `task:${tool_call_id}` instead of
+`subagent_type` so parallel same-type subagents no longer share `messages` and
+`toolCalls` streams. Closes #560.
--- a/libs/deepagents/src/stream.test.ts
+++ b/libs/deepagents/src/stream.test.ts
@@ -322,6 +322,91 @@ describe("streamEvents", () => {
     }
   });
 
+  it("isolates toolCalls for parallel same-type subagent invocations", async () => {
+    const transformer = createSubagentTransformer([])();
+    const projection = transformer.init();
+    const iterator = projection.subagents[Symbol.asyncIterator]();
+
+    const nextA = iterator.next();
+    transformer.process(
+      makeToolEvent([], {
+        event: "tool-started",
+        tool_name: "task",
+        tool_call_id: "task-a",
+        input: {
+          description: "Task A",
+          subagent_type: "researcher",
+        },
+      }),
+    );
+    const subA = (await nextA).value;
+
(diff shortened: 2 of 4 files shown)


christian-bromann:
Added test in #598 which also resolves this bug.

PR #565, opened 2026-06-01
fix(node-vfs): safely rewrite absolute execute paths

### LINKED ISSUE #259: VfsSandbox: execute tool doesn't respect sandbox working directory on macOS
## Summary

  The `execute` tool runs commands in the host's root directory instead of the sandbox directory, causing file system operations to fail with "Read-only file system" errors on macOS.

  ## Steps to Reproduce

  1. Create a VfsSandbox with some initial files
  2. Ask the agent to copy a file from path A to path B using `execute` tool (e.g., `cp /a.txt /b.txt`)
  3. The agent will try to execute the command in the host's root directory instead of the sandbox directory

  ## Expected Behavior

  The command should execute within the sandbox directory, operating on files in the virtual file system.

  ## Actual Behavior

  Error: Read-only file system

  The command executes in the host's root directory (`/`) instead of the sandbox working directory, which is read-only.

  ## Related Issues

  https://github.com/langchain-ai/deepagentsjs/issues/203

  The fix for #203 addressed `ls`, `read_file`, and `write_file` tools, which now work correctly. However, the `execute` tool still has the same issue and was not fixed.

  ## Environment

  - `@langchain/node-vfs`: latest
  - `deepagents`: "^1.8.0"
  - OS: macOS (Darwin 24.5.0)

hntrl:
## Summary

Fixes #259.

This change fixes `@langchain/node-vfs` command execution behavior for absolute paths in `execute()`. The previous implementation only rewrote absolute paths when the top-level path already existed in the VFS root, which caused commands like `cp /a.txt /b.txt` and redirections to `/b.txt` to target host root paths and fail with read-only filesystem errors.

The rewrite behavior is now sandbox-first for workspace paths, while still preserving host system absolute paths for known roots unless those roots are intentionally shadowed by VFS content.

### `@langchain/node-vfs`

- Refactored absolute path rewriting in `VfsSandbox#rewriteVfsPaths` to:
  - Rewrite absolute paths whose root exists in VFS.
  - Rewrite unknown absolute roots to the temp execution workspace.
  - Preserve known host system roots through an explicit allowlist unless shadowed by VFS.
- Moved host absolute root allowlist into `src/const.ts` and imported it from `sandbox.ts`.
- Added integration tests for:
  - `cp /a.txt /b.txt` destination rewriting.
  - Absolute redirection targets with an empty workspace.
  - Host absolute path passthrough (`test -x /bin/sh`).
  - VFS precedence when allowlisted roots are shadowed (for example `/tmp/...`).

### DIFF
--- a/.changeset/vfs-absolute-exec-paths.md
+++ b/.changeset/vfs-absolute-exec-paths.md
@@ -1,0 +1,9 @@
+---
+"@langchain/node-vfs": patch
+---
+
+fix(node-vfs): safely rewrite absolute execute() paths to the sandbox workspace
+
+Update VfsSandbox command rewriting so absolute paths in `execute()` map to the temp execution workspace by default, while preserving known host system roots unless they are explicitly shadowed by VFS entries. This fixes failures like `cp /a.txt /b.txt` and `echo ... > /b.txt`.
+
+Add integration coverage for absolute copy targets, absolute redirection targets, host path passthrough (`/bin/sh`), and VFS precedence when an allowlisted host root (such as `/tmp`) is shadowed in the workspace.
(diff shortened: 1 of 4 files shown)


hntrl:
Closing in favor of a broader architectural change. We should remove execute() from @langchain/node-vfs rather than continue hardening path rewriting on top of host shell execution semantics.

PR #575, opened 2026-06-02
refactor(node-vfs): make VfsSandbox filesystem-only

### LINKED ISSUE #259: VfsSandbox: execute tool doesn't respect sandbox working directory on macOS

hntrl:
## Summary

This PR removes command-execution semantics from `@langchain/node-vfs` and aligns `VfsSandbox` with `BackendProtocolV2` as a filesystem-only backend.

The provider no longer uses sandbox execution abstractions and now performs core file operations directly against the in-memory VFS.

### Protocol alignment

- `VfsSandbox` now implements `BackendProtocolV2` directly instead of extending sandbox/base-execution abstractions.
- Removed command execution from the provider surface (`execute` is no longer part of this backend implementation).
- Removed `timeout` from `VfsSandboxOptions` and corresponding docs.

### Runtime behavior

- Added native backend implementations for:
  - `read`
  - `readRaw`
  - `write`
  - `edit`
  - `ls`
  - `grep`
  - `glob`
- Kept upload/download support and added path confinement handling for invalid traversal paths.
- Continued in-memory VFS-only operation (no temp-dir shell sync path).

### Tests and docs

- Updated unit and integration tests to remove execute-focused assertions and validate filesystem behavior.
- Updated package docs to describe `BackendProtocolV2` usage and removed execute/timeout references.
- Added changeset for `@langchain/node-vfs`.


libs/providers/node-vfs/README.md
```diff
@@ -78,12 +78,13 @@ const sandbox = await VfsSandbox.create({
 
 ##### `sandbox.execute(command)`
```
christian-bromann: I would suggest to just remove this section entirely?

libs/providers/node-vfs/src/types.ts
```diff
@@ -48,7 +48,10 @@ export interface VfsSandboxOptions {
   initialFiles?: Record<string, string | Uint8Array>;
 
   /**
-   * Command timeout in milliseconds.
+   * Backward-compatible timeout option.
+   *
+   * Retained for API compatibility; currently unused because this provider
+   * does not execute shell commands.
```
christian-bromann: Should we deprecate this?

christian-bromann (dismissed):
👍

PR #585, opened 2026-06-05
feat(quickjs): implement interpreter libraries for deterministic workflows with composable code libraries

colifran:
### Summary

Adds interpreter libraries to @langchain/quickjs which are pre-loaded modules that agents can import inside the QuickJS sandbox. Libraries bundle tools, dependencies, and documentation behind a simple module interface, with progressive disclosure (agent sees a one-liner; full docs load on demand via readFile).

Ships swarm as the first built-in library: table-based parallel fan-out with auto-batching, structured output, and context isolation. Swarm subagents now support middleware configuration, with patch-tool-calls and anthropic prompt caching injected by default to help with cost control.

Includes an example (examples/repl/interpreter-libraries/) that wires up swarm with a custom evaluator library to run a multi-pass tech stack comparison across three specialized subagents.

Also fixes two pre-existing CI failures unrelated to this feature:
- stream.test.ts asserted tool output text against sub.messages, which only yields AI messages — fixed to check actual AI message content
- Evicted tool results were written without a file extension, so read_file treated them as binary and produced document blocks the Anthropic API rejects. Added .txt extension

### Tests

- Unit tests for swarm table operations including glob parsing (newline-separated paths, no-match/error status strings)
- Unit tests for library loading (loadLibrary)
- Updated eviction path assertions across fs.test.ts, state.test.ts, store.test.ts, composite.test.ts
- Integration tested via examples/repl/interpreter-libraries/run.ts (multi-library agent with swarm + custom evaluator)


libs/deepagents/src/agent.ts
```diff
@@ -316,6 +317,38 @@ export function createDeepAgent<
...
+  for (const mw of customMiddleware) {
+    if (mw.name === "CodeInterpreterMiddleware" && "_subagentPoolRefs" in mw) {
+      const refs = (mw as Record<string, unknown>)
+        ._subagentPoolRefs as SubagentPoolRef[];
+      for (const ref of refs) {
+        ref.current = poolData;
+      }
+    }
+  }
+
```
hntrl, line 351: hmm is there a better way for us to expose subagent specs to CodeInterpreterMiddleware? I think its pretty bad that we have to do this

hntrl: I also think this technically breaks across threads

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -637,7 +637,8 @@ function createTaskTool(options: {
         },
       };
       const result = (await subagent.invoke(
-        subagentState,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any -- langgraph's InvokeStateParameter union includes Command which doesn't apply here
+        subagentState as any,
```
hntrl, line 641: is this something to fix upstream?

libs/providers/quickjs/src/transform.ts
```diff
@@ -142,6 +149,68 @@ export function transformForEval(code: string): string {
   return s.toString();
 }
 
+/**
+ * Convert a static ImportDeclaration AST node to a dynamic `await import()`.
+ */
+function importToAwaitImport(node: AcornNode): string {
```
hntrl, line 155: should our guidance be to just have `await import` in libraries

hntrl (commented):
some Q's about implementation

libs/providers/quickjs/src/transform.ts
colifran: maybe but it doesn't hurt as a safe guard in case the agent goes rogue

PR #587, opened 2026-06-09
feat(deepagents): implement default middleware systemprompt override

Oscar-Umana:
## Summary

Adds a `middlewareSystemPrompts` option to `createDeepAgent` so callers can override system prompts injected by the built-in middleware stack.

Supported overrides:

- `todoList`
- `filesystem`
- `subagent`
- `asyncSubagent`

The new `DeepAgentMiddlewareSystemPrompts` interface is exported from the package entrypoint for consumers who want to type config objects directly.

## Changes

- Adds `middlewareSystemPrompts?: DeepAgentMiddlewareSystemPrompts` to `CreateDeepAgentParams`.
- Wires prompt overrides into the built-in root middleware stack.
- Applies the same todo/filesystem prompt overrides to normalized subagent middleware.
- Supports disabling the subagent task prompt with `subagent: null`, matching existing `createSubAgentMiddleware` behavior.
- Adds runtime coverage that verifies custom middleware prompts replace the defaults.
- Adds a type test confirming the new param is accepted by `createDeepAgent`.

## Validation

- `pnpm --dir libs/deepagents exec vitest run src/agent.test.ts --config vitest.config.ts --typecheck.enabled=false`
- Editor diagnostics pass for the edited files.

### DIFF
--- a/libs/deepagents/src/agent.test-d.ts
+++ b/libs/deepagents/src/agent.test-d.ts
@@ -93,6 +93,23 @@ describe("createDeepAgent types", () => {
     });
   });
 
+  it("should allow default middleware system prompt overrides", () => {
+    createDeepAgent({
+      middlewareSystemPrompts: {
+        todoList: "Use todos sparingly.",
+        filesystem: "Use filesystem tools carefully.",
+        subagent: "Delegate independent work.",
+        asyncSubagent: "Launch async work only when requested.",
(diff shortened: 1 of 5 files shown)


Oscar-Umana:
Hi @christian-bromann what would be the process to get some reviews on this PR?

PR #590, opened 2026-06-10
chore(quickjs): remove skills backend and associated plumbing

colifran:
### Summary

Removes skills backend and associated plumbing from code interpreter middleware and options


hntrl (approved):

luizzappa:
@colifran, @hntrl  After #590, what is the intended replacement for importing executable scripts bundled with skills?

Previously, a skill could declare a module: entrypoint and its script could be loaded in the QuickJS REPL with:
```js
await import("@/skills/<name>");
```

Now `skillsBackend` and the module loader have been removed, and the same import fails with:
```
ReferenceError: could not load module '@/skills/<name>'
```

Are skill scripts expected to be exposed manually as PTC tools now, copied into each eval call, or is another loading mechanism planned?

They do not appear to be loaded automatically. Also, `deepagents@1.13.2` still parses the module: field and adds the old `await import("@/skills/<name>")` instruction to the system prompt, even though `@langchain/quickjs@0.6.2` can no longer resolve it. Is that leftover behavior a bug?

It would be helpful to document a migration path because **PTC tools are not always equivalent to reusable JS/TS modules containing multiple functions and shared logic**.

PR #592, opened 2026-06-10
feat(quickjs): implement default task primitive in code interpreter for programmatic subagent calling

colifran:
### Summary

This PR adds a default task primitive to the code interpreter that allows the agent to natively write code to programmatically dispatch subagents. The task primitive supports passing response schemas for dynamic structured output that helps with more complex orchestration patterns.

This PR also removes the swarm task tool in favor of the task primitive.


libs/providers/quickjs/src/index.ts
```diff
@@ -32,16 +33,16 @@ export {
...
 export {
-  createSwarmTaskTool,
+  SubagentDispatcher,
   VariantCache,
-  type SwarmTaskToolOptions,
-  type SwarmSubAgent,
-  type SwarmTaskMode,
-} from "./tools/swarm-task.js";
+  validateResponseSchema,
+} from "./subagent-dispatch.js";
```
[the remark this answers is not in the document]
colifran: Expected breakage.

libs/providers/quickjs/src/semaphore.ts
hntrl, line 1: do we just want to use something like p-queue instead?

libs/providers/quickjs/src/session.ts
```diff
@@ -562,4 +582,127 @@ export class ReplSession {
...
+          }
+          promise.settled.then(context.runtime.executePendingJobs);
+        })();
+
+        return promise.handle;
+      },
+    );
+
+    context.setProp(context.global, "subagent", hostFn);
+    hostFn.dispose();
```
hntrl, line 721: disposing here works?

libs/providers/quickjs/src/subagent-dispatch.ts
```diff
@@ -0,0 +1,360 @@
...
+const SCHEMA_MAX_PROPERTIES = 32;
+
+/**
+ * TTL cache for dynamically compiled subagent variants.
+ *
+ * Stores runnables keyed by a string identifier. Entries expire after
+ * `ttlMs` of inactivity. When the cache reaches `maxEntries`, the
+ * least-recently-used entry is evicted.
+ */
+export class VariantCache {
```
hntrl: do we need this to be a TTL cache? I used it in python bc its easy, but I wonder if we just take the simpler road and just have a WeakMap or something like that

colifran: was just trying to stay aligned but i think ttl is too much. i think we could just have the cache exist during the lifetime of an eval

libs/providers/quickjs/src/session.ts
colifran: i'll double check

libs/providers/quickjs/src/semaphore.ts
colifran: meh. any reason not to just keep semaphore?

libs/providers/quickjs/src/subagent-dispatch.ts
colifran: honestly i think a map with a maximum number of entries is the simplest. they preserve insertion order so we can implement a really simple LRU by doing a delete + set per access for comparison we're currently doing a full O(n) sweep on each entry to evict based on ttl plus another O(n) scan for LRU. this solution is just two O(1) ops. I think this is about as simple as we can get.

libs/providers/quickjs/src/session.ts
colifran: we do it for every other setProp in the file

libs/providers/quickjs/src/semaphore.ts
hntrl: cause it sits as kind of like an awkward port of semaphore in python. p-queue is better tested and we have history w/ it

colifran: updated.

libs/providers/quickjs/src/middleware.ts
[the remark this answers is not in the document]
colifran: This is by design and mirrors how PTC already works. The repl calls host functions that invoke tools directly by bypassing the parent's tool node.

A few things the analysis misses: the task tool isn't typically HITL-gated on the parent side, so routing through ToolNode wouldn't actually add a permission check. Subagents' own interruptOn middleware still fires through this path since createSubAgent compiles it in at build time. And the eval tool call itself is the natural permission boundary, which the system prompt documents in the "Approval model" section. Defaulting concurrency to 0 would silently break the feature for anyone who configured subagents with the code interpreter middleware.

libs/providers/quickjs/src/session.ts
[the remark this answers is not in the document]
colifran: False positive, other side of the same call path. The session validates input shape and types because that's the FFI boundary contract. Authorization belongs at the subagent level (via compiled interruptOn middleware) and at the eval tool level (via the parent's interruptOn config), not here.

The subagent type allowlist suggestion is redundant since only developer-registered subagent names are resolvable. Calling task({ subagentType: "anything-else" }) already throws "Unknown subagent type."

libs/deepagents/src/agent.ts
```diff
@@ -470,6 +479,9 @@ export function createDeepAgent<
     ] as const,
   }).withConfig({
     recursionLimit: 10_000,
+    configurable: {
+      [SUBAGENT_SPECS_CONFIG_KEY]: subagentSpecsPayload,
+    },
```
hntrl:
```suggestion
```

libs/deepagents/src/agent.ts
```diff
@@ -346,6 +350,11 @@ export function createDeepAgent<
     createPatchToolCallsMiddleware(),
   ] as const;
 
+  const subagentSpecsPayload = buildSubagentSpecsPayload(inlineSubagents, {
+    model,
+    tools: effectiveTools,
+  });
+
```
hntrl:
```suggestion
```

libs/deepagents/src/agent.ts
```diff
@@ -47,6 +47,10 @@ import type {
   SupportedResponseFormat,
 } from "./types.js";
 import { createSubagentTransformer } from "./stream.js";
+import {
+  SUBAGENT_SPECS_CONFIG_KEY,
+  buildSubagentSpecsPayload,
+} from "./middleware/subagents.js";
```
hntrl:
```suggestion
```

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -24,6 +24,22 @@ import { FilesystemPermission } from "../permissions/types.js";
 
 export type { AgentMiddleware };
 
+/**
+ * Config key for subagent specs in LangGraph configurable.
+ *
+ * Set by `createDeepAgent`, read by code interpreter middleware.
+ */
+export const SUBAGENT_SPECS_CONFIG_KEY = "__deepagents_subagent_specs__";
+
```
hntrl:
```suggestion
```

libs/deepagents/src/index.ts
```diff
@@ -182,3 +182,12 @@ export {
   adaptBackendProtocol,
   adaptSandboxProtocol,
 } from "./backends/index.js";
+
+export {
+  SUBAGENT_SPECS_CONFIG_KEY,
```
hntrl:
```suggestion
```

libs/deepagents/src/index.ts
```diff
@@ -182,3 +182,12 @@ export {
...
+  SUBAGENT_RESPONSE_FORMAT_CONFIG_KEY,
+  buildSubagentSpecsPayload,
+  type SubagentSpecEntry,
+  type SubagentSpecsPayload,
```
hntrl:
```suggestion
```

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -744,6 +789,103 @@ export interface SubAgentMiddlewareOptions {
...
+          tools: agent.tools ?? defaults.tools,
+        },
+        runnableBacked: false,
+      });
+    }
+  }
+
+  return { subagents: entries };
+}
+
```
hntrl:
```suggestion
```

hntrl (dismissed):

hntrl (dismissed):

.changeset/ready-banks-cry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@langchain/quickjs": minor
+"deepagents": minor
```
hntrl:
```suggestion
"deepagents": patch
```

hntrl (dismissed):

PR #599, opened 2026-06-15
feat: add interrupt mode to FilesystemPermission

aolsenjazz:
## Description
Ports the Python `interrupt_on` filesystem permission feature to deepagents.js. `FilesystemPermission` rules now accept `mode: "interrupt"`, which pauses a matching filesystem tool call for human approval before it runs — the path-scoped counterpart to `interruptOn`. Exact-path tools (`read_file`/`write_file`/`edit_file`) interrupt only when the target path matches a rule; bulk tools (`ls`/`glob`/`grep`) interrupt when their search subtree could overlap a rule's anchor. Since the upstream JS `humanInTheLoopMiddleware` has no per-call `when` predicate, this adds a small path-gated middleware that emits the same HITL request/response shape, so resuming works identically. Tools already covered by `interruptOn` are excluded so the user's config wins.

## Release Note
Filesystem permission rules now support `mode: "interrupt"` to pause matching file operations for human approval.

## Test Plan
- [ ] Verify `read_file` on an interrupt-matched path pauses with `__interrupt__`, an unrelated path does not, and resuming with an approve decision proceeds.

Made by [Open SWE](https://openswe.vercel.app)

### DIFF
--- a/.changeset/fs-permission-interrupt-mode.md
+++ b/.changeset/fs-permission-interrupt-mode.md
@@ -1,0 +1,13 @@
+---
+"deepagents": minor
+---
+
+Add `interrupt` mode to `FilesystemPermission`
+
+Filesystem permission rules now accept `mode: "interrupt"`, which pauses a
+matching filesystem tool call for human approval before it runs — the
+path-scoped counterpart to `interruptOn`. Exact-path tools (`read_file`,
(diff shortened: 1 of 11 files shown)


aolsenjazz:
Closing this as duplicate of https://github.com/langchain-ai/deepagentsjs/pull/589

PR #600, opened 2026-06-15
feat(vercel): add Vercel Sandbox provider

scotttrinh:
Add first-party support for Vercel Sandbox as a sandbox provider, mostly following the Deno and Daytona providers.

### DIFF
--- a/libs/providers/vercel/tsconfig.json
+++ b/libs/providers/vercel/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "extends": "../../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "dist"
+  },
+  "include": ["src/**/*.ts", "src/*.ts"],
+  "exclude": ["node_modules", "dist"]
+}
--- a/libs/providers/vercel/tsconfig.json
+++ b/libs/providers/vercel/tsconfig.json
@@ -3,6 +3,6 @@
   "compilerOptions": {
     "outDir": "dist"
   },
-  "include": ["src/**/*.ts", "src/*.ts"],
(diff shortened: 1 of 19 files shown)


libs/providers/vercel/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "dist"
+  },
+  "include": ["src/**/*.ts", "src/*.ts"],
```
christian-bromann: This seems to be a redundant pair
```suggestion
  "include": ["src/**/*.ts"],
```

christian-bromann (commented):
Thank you for the contribution 🙏  Some comments:

- can we add `vercel` as scope to `.github/workflows/pr_lint.yml`
- can we add a changeset
- can we propagate a Vercel secret in `.github/workflows/ci.yml`, I assume this must be `VERCEL_OIDC_TOKEN` / `VERCEL_TOKEN` / `VERCEL_PROJECT_ID` / `VERCEL_TEAM_ID` (I can add the action secrets)
- for completion purposes, add an example in `examples/sandbox`

libs/providers/vercel/tsconfig.json
scotttrinh: Ha! That's what I get for copy-pasting the config. Looks like Deno (and maybe some others?) have this overlapping glob! I'll fix it for the one I'm add here, but probably worth auditing the others and pushing up a small fix since you uncovered this.

christian-bromann (approved):
LGTM 👍

PR #606, opened 2026-06-18
chore(quickjs): disallow task as a configurable ptc tool

colifran:
### Summary

The code interpreter now comes with a global `task` tool out of the box. As a result, we should disallow specifying the `task` tool for ptc. This PR implements logic that forbids the task tool as in ptc and throws with a  detailed error message if found. This PR also updates the RLM agent example and READMEs.

### Tests

Unit tests validating expected behavior for task by name and tool instance with "task" name.

### DIFF
--- a/.changeset/fiery-parrots-wonder.md
+++ b/.changeset/fiery-parrots-wonder.md
@@ -1,0 +1,5 @@
+---
+"@langchain/quickjs": minor
+---
+
+chore(quickjs): disallow task as a configurable ptc tool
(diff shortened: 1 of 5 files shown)


.changeset/fiery-parrots-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+"@langchain/quickjs": minor
```
hntrl, line 2:
```suggestion
"@langchain/quickjs": patch
```

hntrl (approved):

.changeset/fiery-parrots-wonder.md
colifran: agreed to minor offline

PR #608, opened 2026-06-20
fix(deepagents): forward subagent results as text

aolsenjazz:
## Summary

A deep agent that delegates to a subagent 400s when that subagent uses an Anthropic server-side tool (web search, web fetch, code execution, MCP). The subagent's final-message content blocks e.g. server_tool_use/..._tool_result were forwarded verbatim as the task ToolMessage content. On the parent agent's next request they serialize into a tool_result, which Anthropic rejects:

`400 invalid_request_error
messages.N.content.0.tool_result.content.0: Input tag 'server_tool_use' found using 'type'
does not match any of the expected tags: 'document','image','search_result','text','tool_reference'`

## Solution

Now matches the Python impl: forward only the subagent's text answer. returnCommandWithStateUpdate now walks back to the last AIMessage with non-empty text and uses that text (a string) as the ToolMessage content.

- Confirmed the Python deepagents impl is not affected (it forwards .text); this aligns JS with Python.
- Reproduced the 400 across {web_search, code_execution, web_fetch} × {claude-haiku-4-5, claude-sonnet-4-6} on the latest published packages; all green after the fix.

## Tests
- New: subagent server-tool blocks are filtered from forwarded content.
- New: walk-back uses the last non-empty AIMessage, skipping a trailing empty message.
- Updated the #239/#245 tests to assert the (now string) ToolMessage content.
- No regressions in subagent-related tests in the evals suite

### DIFF
--- a/libs/deepagents/src/middleware/subagents.ts
+++ b/libs/deepagents/src/middleware/subagents.ts
@@ -464,11 +464,7 @@ function returnCommandWithStateUpdate(
       const text =
         typeof message.content === "string"
           ? message.content.trim()
-          : message.content
-              .filter((block) => block.type === "text")
-              .map((block) => ("text" in block ? block.text : ""))
-              .join("\n")
-              .trim();
+          : message.text?.trim() ?? "";
       if (text) {
         content = text;
         break;
--- a/libs/deepagents/src/middleware/subagents.ts
+++ b/libs/deepagents/src/middleware/subagents.ts
@@ -464,7 +464,7 @@ function returnCommandWithStateUpdate(
       const text =
         typeof message.content === "string"
           ? message.content.trim()
-          : message.text?.trim() ?? "";
+          : (message.text?.trim() ?? "");
       if (text) {
         content = text;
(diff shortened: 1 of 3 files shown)


libs/deepagents/src/middleware/subagents.ts
```diff
@@ -452,16 +452,26 @@ function returnCommandWithStateUpdate(
...
+      const message = messages[i];
+      if (!message || !AIMessage.isInstance(message)) continue;
+      const text =
+        typeof message.content === "string"
+          ? message.content.trim()
+          : message.content
+              .filter((block) => block.type === "text")
+              .map((block) => ("text" in block ? block.text : ""))
+              .join("\n")
+              .trim();
```
christian-bromann: If `message` is a `BaseMessage` we should be able to just do:
```suggestion
          : message.text?.trim() ?? "";
```

aolsenjazz: Oh yeah nice. updating

PR #618, opened 2026-06-23
feat(deepagents): add registry types, items, builders

aolsenjazz:
## Summary
Adds the registry items to be consumed by all of this projects' system:
- **Providers**
  - OpenAI
  - Anthropic
  - Google
  - Fireworks
- **Frameworks**
  - React + Vite
  - Hono
  - Next
  - Nuxt
  - Deno

Adds placeholders for the repos themselves

## Callouts
- The framework metadata in framework configs (particularly thinking of file/dir paths) should map 1:1 with your cookbook repos
- "builder functions" are effectively just a thin typing layer right now - likely more substantive in the future

### DIFF
--- a/libs/create-deepagent/registry/frameworks/hono.ts
+++ b/libs/create-deepagent/registry/frameworks/hono.ts
@@ -1,0 +1,32 @@
+import {
+  createFramework,
+  ProviderAwareFile,
+} from "../../src/registry/framework.js";
+
+const ENV_D_TS: ProviderAwareFile = {
+  path: "worker/env.d.ts",
+  getContent: ({ providerConfig }) => {
+    const envVars = providerConfig.env
+      .map((e) => `  ${e.name}: string;`)
+      .join("\n");
+
+    return `interface Env {
+  ASSETS: Fetcher;
+  SESSIONS: DurableObjectNamespace;
+${envVars}
+  [key: string]: string | undefined;
+}
+`;
+  },
+};
+
+export const hono = createFramework({
+  id: "hono",
+  title: "Hono",
+  defaultProjectName: "hono-deepagents",
+  frameworkDir: "hono",
+  envFilePath: ".env",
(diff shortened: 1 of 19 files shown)


libs/create-deepagent/registry/frameworks/hono.ts
```diff
@@ -0,0 +1,32 @@
+import {
+  createFramework,
+  ProviderAwareFile,
+} from "../../src/registry/framework.js";
+
+const ENV_D_TS: ProviderAwareFile = {
+  path: "worker/env.d.ts",
```
christian-bromann, line 7: What does `env.d.ts` contain?

libs/create-deepagent/registry/providers/fireworks.ts
```diff
@@ -0,0 +1,15 @@
+import { createProvider } from "../../src/registry/provider.js";
+
+export const fireworks = createProvider({
+  id: "fireworks",
+  title: "Fireworks",
+  defaultModel: "accounts/fireworks/models/glm-5p1",
+  dependencies: ["@langchain/community"],
```
christian-bromann: we should use https://github.com/langchain-ai/langchainjs/tree/main/libs/providers/langchain-fireworks

christian-bromann (approved):
LGTM 👍

libs/create-deepagent/registry/providers/fireworks.ts
aolsenjazz: Oop yes that should be `@langchain/fireworks` thanks

libs/create-deepagent/registry/frameworks/hono.ts
aolsenjazz: Provider env vars. I'd like to keep env types rigorous - using this as a POC and to test out the `ProviderAwareFile` API before I commit to adding for all (or scrap)

PR #620, opened 2026-06-23
feat(deepagents): Preflight filesystem and project name checks

aolsenjazz:
## Summary
Add preflight checks prior to project bootstrap:
- Ensure valid project name
- Ensure empty dir, prompts user to overwrite otherwise
  - `--force` bypasses this check
- Ensures parent dir is writeable

## Testing
Manually tested all preflight code paths locally

### DIFF
--- a/libs/create-deepagent/src/utils/validate.ts
+++ b/libs/create-deepagent/src/utils/validate.ts
@@ -5,7 +5,7 @@ export async function dirExists(dir: string): Promise<boolean> {
   try {
     const stat = await fs.promises.stat(dir);
     return stat.isDirectory();
-  } catch (e) {
+  } catch {
     return false;
   }
 }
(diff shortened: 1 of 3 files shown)


libs/create-deepagent/src/utils/validate.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+/** Ensure a directory exists, creating it if necessary. */
+export async function makeDir(dir: string): Promise<void> {
+  await fs.promises.mkdir(dir, { recursive: true });
+}
+
+/** Validate a npm package name. Returns an error message or null if valid. */
+export function validatePkgName(name: string): string | null {
```
christian-bromann, line 35: should we have unit tests for this to consolidate what is a valid package name?

christian-bromann (approved):
One comment, I am ok to defer unit tests to a later point

LGTM 👍

libs/create-deepagent/src/utils/validate.ts
aolsenjazz: Definitely - will put up some unit tests once more testable units of work

PR #623, opened 2026-06-24
feat(deepagents): Copy project and write files

aolsenjazz:
### Housekeeping
- add frameworks to build artifacts
- Move `ProviderAwareFile` to a better parent module
- Add `chatModelClass` to the provider schema, needed for writing `module.ts`
- `provider.dependencies` -> `provider.dependency` not sure why I thought it was a good idea to make this an array

### Feature work:
- copy the select project after preflight
- write files (`model.ts`, `.env`, and custom files for now)

## Callouts
- Up next is modifying package json and wrapping the file mutations flow
- Then wiring things up in the sample projects
-


libs/create-deepagent/src/registry/files/model.ts
```diff
@@ -0,0 +1,36 @@
...
+        dependency,
+        defaultModel,
+        coordinatorModelConfig,
+      } = providerConfig;
+      const pkg = dependency;
+      const extra = coordinatorModelConfig
+        ? `,\n  ${coordinatorModelConfig}`
+        : "";
+
+      return `import { ${chatModelClass} } from "${pkg}";
```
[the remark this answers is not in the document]
aolsenjazz: Yup that comes later

libs/create-deepagent/src/registry/framework.ts
```diff
@@ -31,5 +26,13 @@ export interface FrameworkConfig<T extends string = string> {
 export function createFramework<T extends string>(
   config: FrameworkConfig<T>,
 ): FrameworkConfig<T> {
-  return config;
+  const defaultFiles = [
+    createModelFile(config.agentPath),
```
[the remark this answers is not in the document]
aolsenjazz: Yup that also comes later

libs/create-deepagent/src/commands/create/index.ts
```diff
@@ -55,7 +63,57 @@ function mergeOptions(cliOptions: CLIOptions, tuiOptions: TUIOptions) {
...
+    await writeFile(path.join(projectPath, file.path), content);
+  }
+
+  s.stop();
+}
+
+/**
+ * Create an env file with values the user fills in.
+ */
+export function createEnvFile(
```
aolsenjazz: Don't love putting this here, but it needs to have access to the full TUI config. open to other ideas

libs/create-deepagent/registry/providers/openai.ts
```diff
@@ -4,7 +4,9 @@ export const openai = createProvider({
   id: "openai",
   title: "OpenAI",
   defaultModel: "gpt-5.4-mini",
-  dependencies: ["@langchain/openai"],
+  chatModelClass: "ChatOpenAI",
```
christian-bromann: I wonder if we can leverage initChatModel here since we already maintain a map for model names to chat instances in [langchain](https://github.com/GAFFWC/langchainjs/blob/81204fbce1e0b7cce6f42c9d318f3d703adf7007/libs/langchain/src/chat_models/universal.ts#L54-L136)

aolsenjazz: Would love to hear more about this - I don't know that API

libs/create-deepagent/src/utils/handleError.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it("prints generic messages for a plain Error", () => {
+    handleError(new Error("something broke"));
+    expect(mockError).toHaveBeenCalledWith("something broke");
+  });
+
+  it("prints field errors for a real ZodError", () => {
+    const result = userSchema.safeParse({ name: "", email: "not-an-email" });
+    handleError(result.error);
```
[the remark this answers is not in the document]
aolsenjazz: Open swe this is a unit test, it passes, its ok

libs/create-deepagent/registry/providers/openai.ts
```diff
@@ -3,8 +3,9 @@ import { createProvider } from "../../src/registry/provider.js";
 export const openai = createProvider({
   id: "openai",
   title: "OpenAI",
-  defaultModel: "gpt-5.4-mini",
-  dependencies: ["@langchain/openai"],
+  defaultModel: "openai:gpt-5.4-mini",
+  coordinatorModelConfig: 'reasoning: { effort: "low", summary: "auto" }',
```
christian-bromann: Seems awkward to store this as string 🤔

christian-bromann:
```ts
import { inspect } from "node:util";

const options = {
  foo: "bar",
  bar: 123,
};

const template = `const foo = ${inspect(options, {
  depth: null,
  compact: false,
})}`;

console.log(template);
```

returns:
```
const foo = {
  foo: 'bar',
  bar: 123
}
```

christian-bromann (approved):
Minor nit, LGTM

libs/create-deepagent/src/registry/files/model.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      ...baseProvider,
+      coordinatorModelConfig: {
+        a: { b: { c: { d: "value" } } },
+      },
+    };
+    const content = file.getContent({ providerConfig: provider });
+
+    expect(content).toMatch(
+      new RegExp(
+        String.raw`
```
aolsenjazz: Sort of ugly, but the most human-readable GLM + I could get

christian-bromann: can't we just use `toMatchInlineSnapshot`?

libs/create-deepagent/src/registry/files/model.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        booleanVal: true,
+        nullVal: null,
+        arrayVal: [1, "two", false],
+        objectVal: { nested: "deep" },
+      },
+    };
+    const content = file.getContent({ providerConfig: provider });
+
+    expect(content).toMatch(
+      new RegExp(
```
christian-bromann: same

PR #627, opened 2026-06-25
feat(deepagents): extensible repo address API

aolsenjazz:
## Summary
`frameworkDir` -> `address`
Address is an extensible API that can handle multiple `scheme`s e.g.
```typescript
const ghAddress = {
  scheme: 'github',
  user: 'aolsenjazz',
  repo: 'libsamplerate-js',
};

const localAddress = {
  scheme: 'local',
  path: 'path/to/...',
}
```

There's a good chance that local address become more of a vestigial limb, but I think this more extensible API will be worth

## Shoehorned a couple things too :)
- revert to `ChatModel` impl
- tightly couple Provider impl with `MODEL_PROVIDER_CONFIG`
- Failure mode handling


libs/create-deepagent/src/commands/create/installTemplate.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Download a GitHub template by fetching the repo tarball and extracting it
+ * into the project directory.
+ */
+async function downloadGithubTemplate(
+  address: { owner: string; repo: string; subPath?: string },
+  dest: string,
+): Promise<void> {
+  const execFileAsync = promisify(execFile);
+  const tarballUrl = `https://codeload.github.com/${address.owner}/${address.repo}/tar.gz/HEAD`;
```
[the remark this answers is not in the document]
aolsenjazz: I don't think this is a concern right now. I also don't think the additional maintenance burden is justified. I'm going to leave as-is, feel free to do so with this thread as well

libs/create-deepagent/src/commands/create/transformPackageJson.ts
```diff
@@ -33,7 +33,7 @@ export function transformPackageJson(
   // 3. Inject the selected provider's dependency
   result.dependencies = {
     ...result.dependencies,
-    [config.provider.dependency]: "latest",
+    [config.provider.package]: "latest",
```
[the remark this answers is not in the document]
aolsenjazz: we already talked about htis

libs/create-deepagent/registry/frameworks/hono.ts
```diff
@@ -22,7 +22,12 @@ export const hono = createFramework({
   id: "hono",
   title: "Hono",
   defaultProjectName: "hono-deepagents",
-  frameworkDir: "hono",
+  address: {
+    scheme: "github",
+    owner: "aolsenjazz",
```
christian-bromann, line 27: we can point now to `langchain-ai` , I made the repo public yesterday

libs/create-deepagent/src/commands/create/installTemplate.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    // 2. Extract and validate
+    const extractDir = path.join(tmpDir, "extracted");
+    await makeDir(extractDir);
+
+    await pipeline(
+      Readable.fromWeb(response.body as NodeReadableStream<Uint8Array>),
+      tarExtract({ cwd: extractDir, strip: 1 }),
+    );
```
christian-bromann, line 63: Clean 👌

christian-bromann (approved):
One comment, otherwise LGTM 👍

libs/create-deepagent/src/commands/create/installTemplate.ts
aolsenjazz: 😎

libs/create-deepagent/registry/frameworks/hono.ts
aolsenjazz: Gonna keep pointed at my fork until I have the full matrix (package manager x framework x provider) tested

aolsenjazz: Then will structure some PRs on the `langchain-ai` repo to keep things neat

PR #628, opened 2026-06-25
ci: enforce sha-pinned github actions

jkennedyvz:
## Summary
- add PR linting that rejects external GitHub Actions not pinned to a 40-character commit SHA
- keep existing action refs pinned and align stale version comments with their pinned SHAs
- extend .gitignore coverage for local env, key/cert, Python cache, and virtualenv files

## Testing
- local SHA-pin check over .github/workflows/*.yml and *.yaml
- ruby YAML parse for .github/workflows/*.yml
- git diff --cached --check


christian-bromann (approved):
🙌

PR #643, opened 2026-07-02
fix(deepagents): let grep match an exact file path on State/Store backends

NitsanCohen770:
## Problem

`grepMatchesFromFiles` (the shared grep for `StateBackend` and `StoreBackend`) normalizes its `path` argument via `validatePath`, which **always appends a trailing slash**, then keeps only files whose key starts with that prefix:

```
grep(pattern, "/data/rows.txt")
  → normalizedPath = "/data/rows.txt/"
  → files["/data/rows.txt"].startsWith("/data/rows.txt/")  // false
  → []
```

So grepping a **specific file** — a perfectly normal call, and one that works against shell-backed sandboxes — silently returns zero matches on state/store backends. Silent-empty is the worst failure mode here: an agent concludes "the pattern isn't in the file" and answers wrong, with no error to catch.

## Fix

The filter also accepts an exact file-key match (`` `${fp}/` === normalizedPath ``). Directory-subtree semantics, glob filtering, and the binary-file skip are unchanged.

## Test

Adds `grepMatchesFromFiles` unit coverage (directory subtree, exact file path, exact file without the pattern, glob applied to an exact file path) and a `StoreBackend.grep` regression test for the exact-file case.

## Verification

Hit in production: our tool-result offload handle instructs the model to `grep(pattern, "<offloaded file path>")`; against the store backend every such call returned `[]` for data written moments earlier. `libs/deepagents` vitest suite: 1181 passing locally (the 6 pre-existing `*.int.test.ts` typecheck errors about `@langchain/sandbox-standard-tests` are unrelated and present on `main`).

Adjacent observation (out of scope here): files with text content but an unrecognized extension (e.g. `.jsonl`) get a non-text MIME and are skipped by grep's binary filter even when the path matches — worth a separate look.

### DIFF
--- a/libs/deepagents/src/backends/store.test.ts
+++ b/libs/deepagents/src/backends/store.test.ts
@@ -212,6 +212,20 @@ describe("StoreBackend", () => {
     expect(readAfter.content).not.toContain("foo");
   });
 
+  it("greps an exact file path, not only directory prefixes", async () => {
+    const { runtime } = makeConfig();
+    const backend = new StoreBackend(runtime);
+
+    await backend.write("/data/rows.txt", '{"account":"acct-1"}');
+    await backend.write("/data/other.txt", "acct-1 elsewhere");
+
+    const res = await backend.grep("acct-1", "/data/rows.txt");
+    expect(res.error).toBeUndefined();
+    expect(res.matches).toEqual([
+      { path: "/data/rows.txt", line: 1, text: '{"account":"acct-1"}' },
+    ]);
+  });
+
   it("should handle grep with glob filter", async () => {
     const { runtime } = makeConfig();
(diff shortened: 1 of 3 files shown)


NitsanCohen770:
Closing this — it was superseded upstream by #713, released in 1.12.1, which fixes the same defect more thoroughly: instead of adding an exact-key check inside `grepMatchesFromFiles`, it extracts a shared `filterFilesByPath` helper that treats a file `path` as a single-file search, and applies it to `glob` as well, which this PR did not cover.

Verified against 1.12.2 before closing:

```
grep("…", "/tooldata/catalog.json")  -> matches (exact file path)
grep("…", "/tooldata")               -> matches (directory)
```

Thanks @gethin-langchain for landing it.

PR #644, opened 2026-07-03
fix(deepagents): use file_path in read_file prompt examples

TrueNight:
## Problem

The `read_file` tool schema requires the argument `file_path`, but two built-in prompts illustrate the call with an argument named `path`:

- `READ_FILE_TOOL_DESCRIPTION` in `libs/deepagents/src/middleware/fs.ts` (`read_file(path, limit=...)`, `read_file(path, offset=..., limit=...)`)
- the skills workflow prompt in `libs/deepagents/src/middleware/skills.ts` (`read_file(path, limit=...)`)

Models that copy the example emit a `path` argument and fail schema validation on the very first read:

```
Error invoking tool 'read_file' with kwargs {"path":".../SKILL.md","limit":1000}
Error: Received tool input did not match expected schema
Invalid input: expected string, received undefined
  -> at file_path
```

## Change

Align the four example lines with the actual schema by renaming the example argument `path` → `file_path`. This is a prompt/documentation fix only — the tool schema (`file_path`) is unchanged, so there is no API/behavior change for callers.

## Test

Adds one test under `createFilesystemMiddleware › tools` in `fs.test.ts` that derives the allowed argument names from the `read_file` tool's schema (`schema.toJSONSchema().properties`) and asserts every argument shown in the tool description's `read_file(...)` examples is one of them. It fails if an example ever references an argument the tool does not accept (verified red before this change, green after).

## Verification

- `pnpm build` — passes
- `pnpm --filter deepagents typecheck` — passes
- `pnpm --filter deepagents test:unit` — 1177 passed

A patch changeset is included.

### DIFF
--- a/.changeset/fix-read-file-file-path.md
+++ b/.changeset/fix-read-file-file-path.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+Fix built-in `read_file` prompt examples to use the `file_path` argument name so models stop emitting `path` and failing schema validation.
--- a/libs/deepagents/src/middleware/fs.test.ts
+++ b/libs/deepagents/src/middleware/fs.test.ts
@@ -1048,6 +1048,32 @@ describe("createFilesystemMiddleware", () => {
         }
       }
     });
+
+    it("read_file examples only reference argument names in its schema", () => {
+      const middleware = createFilesystemMiddleware({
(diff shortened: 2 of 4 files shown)


TrueNight:
Hi @hntrl! Could you take a look at this PR when you have a moment? It’s ready for review

Issue #658, opened 2026-07-08
Filesystem tools fail validation on weaker/custom models due to path/file_path parameter mismatch


ghost:
### Describe the bug
When using weaker or custom LLMs (e.g., smaller open-source models or custom endpoints) in `deepagents` and consumers like `openwiki`, the agent immediately crashes with a schema validation error on the first `read_file` call:

### Root Cause
There are two contributing factors to this issue:
1. **Schema Inconsistency**: The `ls` tool defines its directory argument as `path`, while `read_file`, `write_file`, and `edit_file` define their target file argument as `file_path`. Models often generalize tool naming conventions and default to using `path` across all filesystem tools.
2. **Confusing Prompt Examples**: Both `READ_FILE_TOOL_DESCRIPTION` (in `fs.ts`) and the skills system prompt (in `skills.ts`) show documentation examples instructing the model to use `path`, e.g. `read_file(path, limit=100)`. Less capable models strictly copy this behavior and fail the actual Zod schema validation because it expects `file_path`.

### Proposed Solution
We should implement both a prompt correction and a **defensive input normalization** layer so that filesystem tools work reliably across all model tiers.

1. **Prompt Corrections**: Align the documentation examples in `fs.ts` and `skills.ts` to use `file_path` (as initiated in PR #644).
2. **Schema Input Normalization**: Add a `z.preprocess` helper to `read_file`, `write_file`, and `edit_file` schemas to dynamically map `path` to `file_path` if the model passes `path`.

#### Implementation Example:
```typescript
function normalizeFilePathInput(input: unknown): unknown {
  if (
    typeof input === "object" &&
    input !== null &&
    "path" in input &&
    !("file_path" in input)
  ) {
    const { path, ...rest } = input as Record<string, unknown>;
    return { ...rest, file_path: path };
  }
  return input;
}

// Wrap schemas
schema: z.preprocess(normalizeFilePathInput, z.object({
  file_path: z.string().describe("Absolute path to the file to read"),
  // ...
}))

PR #659, opened 2026-07-08
feat(deepagents): implement filesystem state management middleware with atomic update support and testing

ghost:
## Problem
Addresses #658.

When using weaker or custom LLMs (e.g., smaller open-source models) in `deepagents` and consumers like `openwiki`, the agent immediately crashes with a schema validation error on the first filesystem tool call:

This occurs because:
1. `ls` defines its directory argument as `path`, but `read_file`, `write_file`, and `edit_file` expect `file_path`. Models often default to `path` across all filesystem tools.
2. The description prompt examples in `fs.ts` and `skills.ts` show `read_file(path, ...)`, instructing models to use the wrong parameter name.

## Proposed Changes
This PR solves both prompt inconsistencies and schema validation failures with a fully backward-compatible input normalization helper:

1. **Prompt Alignment**: Fixed references in `fs.ts` and `skills.ts` to instruct the model to use `file_path` (aligned with the prompt fixes in #644).
2. **Schema Input Normalization**: Added a `normalizeFilePathInput` preprocessor helper to the `read_file`, `write_file`, and `edit_file` Zod schemas using `z.preprocess`. This dynamically maps the `path` key to `file_path` if the model still passes `path`.

## Verification & Tests
- Added unit tests in `fs.test.ts` to assert that:
  - `path` parameter normalization maps correctly to `file_path` for `read_file`, `write_file`, and `edit_file`.
  - All instruction examples contain only valid schema fields.
- Verified that `zod-to-json-schema` unwraps the `z.preprocess` layer perfectly, preserving the original schema descriptions and definitions sent to the LLM.
- Ran all filesystem tests successfully (`npx vitest run src/middleware/fs.test.ts` passes).

### DIFF
--- a/.changeset/normalize-fs-path-parameter.md
+++ b/.changeset/normalize-fs-path-parameter.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+Fix: Normalize `path` to `file_path` in filesystem tools (`read_file`, `write_file`, and `edit_file`) and align the prompt documentation examples to prevent validation schema failures on weaker/custom models.
--- a/libs/deepagents/src/middleware/fs.test.ts
+++ b/libs/deepagents/src/middleware/fs.test.ts
@@ -1048,6 +1048,58 @@ describe("createFilesystemMiddleware", () => {
         }
       }
     });
+
+    it("filesystem schemas should normalize 'path' parameter to 'file_path'", () => {
+      const middleware = createFilesystemMiddleware({
+        backend: createMockBackend(),
+      });
+
+      const readFileTool = middleware.tools!.find((t: any) => t.name === "read_file") as any;
+      const writeFileTool = middleware.tools!.find((t: any) => t.name === "write_file") as any;
+      const editFileTool = middleware.tools!.find((t: any) => t.name === "edit_file") as any;
+
+      // Verify read_file normalization
+      const parsedRead = readFileTool.schema.parse({ path: "/foo/bar.txt", limit: 10 });
+      expect(parsedRead.file_path).toBe("/foo/bar.txt");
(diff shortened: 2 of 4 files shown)


Kowshik4593:
@colifran  can you verify these changes as this is a blocker for openwiki too! we are facing the same issue in openwiki.

**Context & Why PR #644 is not enough:**
While #644 aligns the prompt examples, weaker/custom models still frequently default to sending `path` instead of `file_path` due to prior training/habits and the fact that `ls` uses `path`.

**What this PR adds:**
1. **Robust Schema Preprocessing**: Added `z.preprocess` normalization to `read_file`, `write_file`, and `edit_file` to dynamically map `path` -> `file_path` under the hood if `file_path` is not provided.
2. **Corrected Prompt Examples**: Renamed `read_file(path, ...)` -> `read_file(file_path, ...)` in `fs.ts` and `skills.ts`.
3. **Changeset**: Added a patch changeset for the `deepagents` package.

**Verification & Safety:**
- Added comprehensive unit tests in `fs.test.ts` to verify parameter mapping works and that prompt examples match valid properties.
- Running the vitest unit tests (`npx vitest run src/middleware/fs.test.ts`) succeeds with all 61 tests passing.
- Confirmed `zod-to-json-schema` unwraps `z.preprocess` cleanly, meaning there is zero impact on the JSON Schema presented to the LLM.
- Resolved code formatting checks (`npx pnpm format` was run).

This is a fully backward-compatible solution that fixes the validation failures for weaker models.

colifran:
@Kowshik4593 nice i was actually just starting to look into the root cause since i agree this is another larger blocking issue. i'll take a look here.

colifran (approved):
Nice - thank you!

PR #691, opened 2026-07-16
fix(deepagents): backend adapter drops route prefixes

colifran:
### Summary

`adaptBackendProtocol` rebuilds the backend into a v2 backend protocol but drops its `routePrefixes`. Since `resolveBackend` runs every backend through the adapter, a resolved `CompositeBackend` was no longer detected as composite. That made the execute tool's guard fail at invocation and disable the execute tool even when filesystem permissions were correctly scoped to composite routes.

This PR fixes this by forwarding `routePrefixes` through `adaptBackendProtocol` instead of dropping them.

### Tests

Added unit tests in backends/utils.test.ts to verify that `routePrefixes` is forwarded through `adaptBackendProtocol` and `adaptSandboxProtocol` and not added for non-composite backends.

### DIFF
--- a/.changeset/dark-squids-bathe.md
+++ b/.changeset/dark-squids-bathe.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+fix(deepagents): backend adapter drops route prefixes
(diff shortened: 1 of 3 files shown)


colifran:
tested locally - need to merge for release.

PR #697, opened 2026-07-20
ci: route integration tests through LangSmith LLM Gateway

christian-bromann:
## Summary
- Add a shared Vitest setup that remaps Anthropic/OpenAI/etc. clients onto the LangSmith LLM Gateway when `LANGSMITH_GATEWAY_KEY` is present.
- Wire that setup into all package int-test Vitest configs.
- Pass `LANGSMITH_GATEWAY_KEY` into the CI integration-test job (falls back to direct provider keys if unset).


colifran (approved):
Nice!

PR #716, opened 2026-07-31
fix: avoid duplicate sandbox read line numbers

### LINKED ISSUE #679: Bug: `read_file` line numbers are duplicated when using a sandbox backend
## Summary

When a file is read through any sandbox backend (`BaseSandbox` / its subclasses in `libs/deepagents/src/backends/sandbox.ts`), the output shows line numbers **twice** on every line, e.g.:

```
     1	     1	import { foo } from "bar";
     2	     2	const x = 1;
```

instead of the expected single column:

```
     1	import { foo } from "bar";
     2	const x = 1;
```

## Root cause

Line numbering is applied in **two** places for sandbox reads:

1. **The backend.** `buildReadCommand()` in `libs/deepagents/src/backends/sandbox.ts` builds an `awk` command that prepends line numbers itself:

   ```js
   awk 'NR >= ${start} && NR <= ${end} { printf "%6d\t%s\n", NR, $0 }' ${quotedPath}
   ```

2. **The middleware.** The shared `read_file` tool in `libs/deepagents/src/middleware/fs.ts` (line ~773) unconditionally re-numbers whatever content the backend returns:

   ```js
   let formatted = formatContentWithLineNumbers(content, offset + 1);
   ```

## Proposed fix

Change the `awk` command in `buildReadCommand()` to emit only the line content:

```js
awk 'NR >= ${start} && NR <= ${end} { printf "%s\n", $0 }' ${quotedPath}
```

mikemikimike:
## Summary

- Stop sandbox backends from adding line-number prefixes to file content.
- Keep line-number formatting in the shared filesystem middleware, so sandbox reads use the same single formatting path as other backends.
- Add regression coverage proving the generated awk command returns raw content.

Closes #679

## Validation

- Sandbox backend tests: 57 passed.
- Deepagents unit suite: 1255 passed, 3 existing Windows `local-shell` failures (working-directory, timeout, and bash integration behavior).
- Targeted oxlint and oxfmt: passed.
- Package typecheck/integration discovery is currently blocked by the existing workspace package `@langchain/sandbox-standard-tests/vitest` not having a generated `dist` export in this checkout.


mikemikimike:
Validation details:

- `pnpm install --frozen-lockfile`: completed successfully.
- `pnpm --filter deepagents exec vitest run src/backends/sandbox.test.ts`: **57 passed**, no type errors.
- `pnpm --filter deepagents test:unit`: **1255 passed / 3 failed**. The three failures are existing Windows `local-shell` behavior tests (working directory, timeout, and bash integration), unrelated to the sandbox backend diff.
- `pnpm exec oxlint libs/deepagents/src/backends/sandbox.ts libs/deepagents/src/backends/sandbox.test.ts`: passed.
- `pnpm exec oxfmt --check ...`: passed.
- Package typecheck and integration discovery are blocked by the existing workspace package `@langchain/sandbox-standard-tests/vitest` missing its generated `dist` export in this checkout; the build itself fails because the repository's tsdown config imports unavailable `unrun`.

The branch is clean at commit `3b11c83`.

PR #723, opened 2026-08-04
fix(deepagents): prevent stack overflow in CompositeBackend grep/glob

thushanth-bengre-langchain:
## Summary

- Fixes `RangeError: Maximum call stack size exceeded` from `CompositeBackend.glob`/`grep` when a broad search (e.g. `**/*.{ts,tsx,...}` at `/`) returns hundreds of thousands of entries. Root cause: results were merged with `push(...entries)`, which passes each entry as a separate function argument and overflows V8's argument-stack limit. Now accumulated with a plain loop.
- Ports the Python SDK's grep match-count cap (`ef591e7`): optional `maxCount` backend param / `max_count` tool arg, `grepMaxCount` middleware option (default 1000, `null` disables), `truncated` flag on `GrepResult`/`GlobResult`, and a truncation note in the grep tool output. `CompositeBackend` splits the budget across routes and OR-propagates `truncated`.

## Backward compatibility

- `truncated` is optional; `maxCount` defaults to unset everywhere. No required changes for existing or new users.

## Test plan

- [x] Regression test: mocked backend returning 200k entries — `glob`/`grep` complete without `RangeError` (verified it fails with the old spread).
- [x] Cap enforcement, `truncated` propagation across composite routes, middleware note rendering.
- [x] `pnpm test` (format + lint + all lib tests) passes.

### DIFF
--- a/libs/deepagents/src/backends/composite.ts
+++ b/libs/deepagents/src/backends/composite.ts
@@ -19,8 +19,28 @@ import type {
   ReadResult,
   WriteResult,
 } from "./protocol.js";
-import { isSandboxBackend, isSandboxProtocol } from "./protocol.js";
+import {
+  applyGrepMaxCount,
+  isSandboxBackend,
+  isSandboxProtocol,
+} from "./protocol.js";
 import { adaptBackendProtocol, adaptSandboxProtocol } from "./utils.js";
+
+/**
+ * Remaining match budget for the next routed grep, or null when uncapped.
+ *
+ * Returns 0 once the cap is already met so callers can short-circuit
+ * remaining routes.
+ */
+function remainingGrepBudget(
+  maxCount: number | null | undefined,
+  collected: number,
+): number | null {
+  if (maxCount == null) {
+    return null;
+  }
+  return Math.max(maxCount - collected, 0);
+}
 
 /**
  * Backend that routes file operations to different backends based on path prefix.
@@ -179,7 +199,11 @@ export class CompositeBackend implements BackendProtocolV2 {
         return defaultResult;
       }
 
-      results.push(...(defaultResult.files || []));
+      // Loop instead of spread: push(...files) passes each entry as a
+      // separate argument and overflows the call stack on huge listings.
+      for (const fi of defaultResult.files || []) {
+        results.push(fi);
+      }
 
       // Add the route itself as a directory (e.g., /memories/)
       for (const [routePrefix] of this.sortedRoutes) {
@@ -229,11 +253,16 @@ export class CompositeBackend implements BackendProtocolV2 {
 
   /**
    * Structured search results or error string for invalid input.
+   *
+   * @param maxCount - Optional total cap on returned matches across all routed
+   *                   backends. When the cap is reached, remaining routes are
+   *                   short-circuited and the result is flagged `truncated: true`.
    */
   async grep(
     pattern: string,
     path: string | null = "/",
     glob: string | null = null,
+    maxCount: number | null = null,
   ): Promise<GrepResult> {
     const searchPath = path || "/";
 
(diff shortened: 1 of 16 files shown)


libs/deepagents/src/backends/composite.ts
```diff
@@ -310,16 +364,22 @@ export class CompositeBackend implements BackendProtocolV2 {
...
 
     // Path doesn't match any specific route - search default and route descendants
+    let truncated = false;
     const defaultResult = await this.default.glob(pattern, path);
     if (defaultResult.error) {
       return defaultResult;
     }
-    results.push(...(defaultResult.files || []));
+    // Loop instead of spread: push(...files) passes each entry as a
+    // separate argument and overflows the call stack on huge result sets.
```
hntrl: let's remove these comments

libs/deepagents/src/backends/composite.ts
```diff
@@ -310,16 +364,22 @@ export class CompositeBackend implements BackendProtocolV2 {
...
+    for (const fi of defaultResult.files || []) {
+      results.push(fi);
+    }
+    truncated = truncated || defaultResult.truncated === true;
```
hntrl: these lines look a little strange. Why do we have a separate truncated marker instead of just reading off of `defaultResult.truncated`?

libs/deepagents/src/backends/protocol.ts
```diff
@@ -71,8 +71,41 @@ export interface GrepMatch {
...
+/**
+ * Enforce a match cap after a backend grep has completed.
+ *
+ * When `maxCount` is set and the result exceeds it, the matches are sliced
+ * to the cap and the result is flagged `truncated: true`.
+ */
+export function applyGrepMaxCount(
+  result: GrepResult,
+  maxCount: number | null | undefined,
+): GrepResult {
```
hntrl: nit: this signature kind of pains me. could we instead have something like `params: { result: GrepResult, maxCount: number | null | defined }` instead of two positional args?

libs/deepagents/src/backends/composite.ts
```diff
@@ -19,9 +19,29 @@ import type {
...
+ */
+function remainingGrepBudget(
+  maxCount: number | null | undefined,
+  collected: number,
+): number | null {
+  if (maxCount == null) {
+    return null;
+  }
+  return Math.max(maxCount - collected, 0);
+}
```
hntrl: looks like this is only used once? could we just have this logic live inline in that case

hntrl (commented):
some Q's

libs/deepagents/src/backends/composite.ts
thushanth-bengre-langchain: This is the merge path — it searches default + all routed backends and ORs truncated across them. Only `defaultResult.truncated` would miss truncation from routed backends.

libs/deepagents/src/middleware/fs.ts
```diff
@@ -1056,7 +1072,9 @@ function createGrepTool(
 
       const resolvedBackend = await resolveBackend(backend, runtime);
       const { pattern, path = "/", glob = null } = input;
-      const result = await resolvedBackend.grep(pattern, path, glob);
+      // A per-call max_count overrides the configured middleware default.
+      const maxCount = input.max_count ?? grepMaxCount;
+      const result = await resolvedBackend.grep(pattern, path, glob, maxCount);
```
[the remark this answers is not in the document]
thushanth-bengre-langchain: Good catch — agreed this is a real issue, but it also exists in the Python implementation (same order of operations: backend cap applied before permission filtering), so JS is currently at parity rather than regressed. The right fix likely needs a design decision (over-fetch vs. pushing path-awareness into backends) that affects both repos.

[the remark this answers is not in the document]
hntrl: we're disregarding this for now. This will require a broader rearchitecture to the protocol to allow for permission invariants while iterating. Will open up a corresponding PRD to capture this

hntrl (approved):

PR #724, opened 2026-08-04
feat(deepagents): add output_mode parameter to the grep tool

gethin-langchain:
## Summary
- The grep tool's description mentioned output_mode, but it was never an actual parameter the tool always returned one hardcoded format.
- Added output_mode (files_with_matches/content/count) as a real schema parameter, using the existing formatGrepMatches helper.

## Test plan
- Added tests in fs.test.ts for files_with_matches and count modes.
- All tests pass.

Note: default output is now sorted by path with no blank lines between file groups (minor formatting change, same info).

### DIFF
--- a/.changeset/grep-output-mode.md
+++ b/.changeset/grep-output-mode.md
@@ -1,0 +1,5 @@
+---
+"deepagents": minor
+---
+
+feat(deepagents): add `output_mode` parameter to the `grep` tool (`files_with_matches` / `content` / `count`)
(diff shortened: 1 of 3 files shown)


.changeset/grep-output-mode.md
```diff
@@ -0,0 +1,5 @@
+---
+"deepagents": minor
```
hntrl:
```suggestion
"deepagents": patch
```

hntrl (dismissed):

hntrl (dismissed):

PR #737, opened 2026-08-10
fix(deepagents): return tool exceptions as error messages

### LINKED ISSUE #655: Unhandled tool exceptions (e.g. fs EACCES) abort the whole agent turn instead of returning a tool error the model can react to
**Packages:** `deepagents` 1.10.x / `deepagents-acp` v0.1.15

## What happens

When a built-in tool raises mid-turn — observed with the search/grep tool walking into a directory the process can't read — the exception propagates out of the tool node and kills the entire turn. Over ACP it surfaces as a JSON-RPC error:

```
Internal error — {"details":"EACCES: permission denied, scandir '/lost+found'"}
```

All context from the turn is lost; the client has no way to continue the session's work.

## Why this matters

The prevailing agent-loop contract (Claude Code / Agent SDK's `is_error: true` tool results, Codex CLI's stringified `tool_output` failures) is that a failed tool call is **fed back to the model as an observation** so it can adapt — retry with different arguments, pick another path. A permission error on one directory is exactly the kind of thing a model self-corrects from; hard-aborting turns a recoverable hiccup into a lost investigation. We hit this in production-style use (driving deepagents-acp from [prismalens](https://github.com/prismalens/prismalens)): a 17-tool-call investigation died on tool call 3 because of one unreadable directory.

We've seen the same abort-the-turn shape from tool-argument schema rejections (the model emitting an argv array where a string was expected), so the general fix is at the tool-execution boundary rather than per-tool.

## Suggested fix

Catch exceptions at the tool-node boundary and return them as an error `ToolMessage` (message + stack line) so the LLM sees the failure as the tool's result. A hard abort could remain for genuinely fatal classes (model/API auth, context overflow), but filesystem and validation errors are model-recoverable.

Related: #653 (the absolute-path escape that produced this particular EACCES) and #654.

mikemikimike:
Fixes #655

## Summary
- Convert exceptions raised by wrapped tools into `ToolMessage` instances with `status: error`.
- Preserve the tool name, call ID, and stack information so the model can react to the failure.
- Cover tool-call execution with filesystem result eviction enabled and disabled.
- Add a patch changeset.

## Validation
- `pnpm --filter deepagents exec vitest run src/middleware/fs.test.ts`
- `pnpm --filter deepagents build`
- `pnpm format:check`
- `pnpm lint`
- Full unit suite: 1271/1274 tests passed; 3 existing Windows local-shell tests fail because they assume Unix commands, and existing integration-test type errors are caused by the missing sandbox test module.

### DIFF
--- a/.changeset/tool-errors-as-messages.md
+++ b/.changeset/tool-errors-as-messages.md
@@ -1,0 +1,5 @@
+---
+"deepagents": patch
+---
+
+fix(deepagents): return tool exceptions as error messages
--- a/libs/deepagents/src/middleware/fs.test.ts
+++ b/libs/deepagents/src/middleware/fs.test.ts
@@ -599,6 +599,38 @@ describe("createFilesystemMiddleware", () => {
       expect(result).toBe(mockMessage);
     });
 
+    it.each([null, 100])(
+      "should convert thrown tool errors into error ToolMessages (limit: %s)",
+      async (toolTokenLimitBeforeEvict) => {
+        const middleware = createFilesystemMiddleware({
+          backend: createMockBackend(),
+          toolTokenLimitBeforeEvict,
+        });
+        const mockHandler = vi
+          .fn()
+          .mockRejectedValue(new Error("permission denied"));
+        const request = {
+          toolCall: { id: "test-id", name: "test_tool" },
+          runtime: { toolCall: { id: "test-id", name: "test_tool" } },
+          state: {},
+          config: {},
+        };
+
+        const result = await middleware.wrapToolCall!(
+          request as any,
+          mockHandler,
+        );
+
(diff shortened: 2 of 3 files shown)


hntrl (commented):
Hi @mikemikimike, our general guidance for handling tool failures is by imposing additional middleware that can appropriately catch tool failures. You can see our rationale for why [here](https://github.com/langchain-ai/deepagents/issues/5356#issuecomment-5257811491)

(see [fault tolerance](https://docs.langchain.com/oss/javascript/deepagents/fault-tolerance#error-handling-strategies), [toolRetryMiddleware](https://reference.langchain.com/javascript/langchain/index/toolRetryMiddleware), [toolErrorMiddleware](https://reference.langchain.com/javascript/langchain/index/toolErrorMiddleware))

PR #739, opened 2026-08-10
fix(deepagents): extract text from content blocks when building the summary

taoche:
## Description

`SummarizationMiddleware`'s `createSummary` falls back to `JSON.stringify(response.content)` when the summarization model's response content is not a string:

```ts
return typeof response.content === "string"
  ? response.content
  : JSON.stringify(response.content);
```

When the summarization model responds with a content block array, the raw blocks are stringified verbatim into the summary `HumanMessage`. This happens in practice when:

- **Thinking/reasoning is enabled** on the summarization model (e.g. Claude on Bedrock with `thinking`): the content is `[{type: "reasoning", reasoning: ..., signature: <opaque multi-KB signature>}, {type: "text", ...}]`, and the reasoning signature lands in the summary.
- **`invoke()` takes the internal streaming path** (a `prefersStreaming` callback handler is attached, e.g. inside `streamEvents`, and `disableStreaming` is not set): the aggregated `AIMessageChunk` content is an array of many small text fragments.

Observed in production: a summary message of ~56KB consisting of an ~8KB opaque reasoning signature plus hundreds of `{"type":"text","text":"..."}` fragments — the "summary" ended up bloating the context it was supposed to compact, and subsequent model calls overflowed.

## Fix

Use the message's `text` accessor for non-string content, which extracts and joins only the text blocks. String content is returned as-is (unchanged behavior).

## Testing

Added a unit test where the summarization model returns reasoning + split text blocks, asserting the summary message contains the joined text and no reasoning payload. All 33 existing tests pass.

### DIFF
--- a/.changeset/summary-text-extraction.md
+++ b/.changeset/summary-text-extraction.md
@@ -3,5 +3,3 @@
 ---
 
 fix(deepagents): extract text from content blocks when building the summary
-
-When the summarization model responds with a content block array (e.g. reasoning/thinking enabled, or chunk aggregation from an internal streaming path), `createSummary` previously `JSON.stringify`'d the raw blocks — reasoning signatures included — into the summary message, bloating it instead of compacting the history. Use the message's `text` accessor to keep only text blocks.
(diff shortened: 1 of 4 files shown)


hntrl (approved):
Thanks for the pr! 🙏

PR #749, opened 2026-08-13
fix(deepagents): exclude summarization state from subagent input/output

### LINKED ISSUE #748: Subagents inherit parent's summarization cutoff, can silently drop their own task message
### Problem
`EXCLUDED_STATE_KEYS` in `subagents.ts` doesn't exclude `_summarizationEvent`/`_summarizationSessionId`, so they leak into every subagent call, forked or not.

```ts
// subagents.ts
const EXCLUDED_STATE_KEYS = [
  "messages", "todos", "structuredResponse", "skillsMetadata", "memoryContents",
] as const;
```

A subagent's `messages` starts as one `HumanMessage`. If the parent already summarized once, its stale `cutoffIndex` (e.g. `40`) leaks in, and `messages.slice(40)` on that 1-message array returns `[]` — the effective view collapses to just the parent's old summary, silently dropping the subagent's actual task.

Same bug in reverse: a subagent's own summarization event can flow back and overwrite the parent's (no custom reducer, last-write-wins), corrupting the parent's own next reconstruction.

### Fix
Add both keys to `EXCLUDED_STATE_KEYS`.

thushanth-bengre-langchain:
Fixes #748 

Excludes `_summarizationEvent`/`_summarizationSessionId` from subagent state input/output. These weren't in `EXCLUDED_STATE_KEYS`, so a stale parent cutoffIndex could silently drop a subagent's task description, and a subagent's own summarization event could overwrite the parent's on return.

### DIFF
--- a/libs/deepagents/src/middleware/subagents.ts
+++ b/libs/deepagents/src/middleware/subagents.ts
@@ -41,16 +41,9 @@ export const DEFAULT_SUBAGENT_PROMPT =
   "In order to complete the objective that the user asks of you, you have access to a number of standard tools.";
 
 /**
- * State keys that are excluded when passing state to subagents and when returning
- * updates from subagents.
- *
- * When returning updates:
- * 1. The messages key is handled explicitly to ensure only the final message is included
- * 2. The todos and structuredResponse keys are excluded as they do not have a defined reducer
- *    and no clear meaning for returning them from a subagent to the main agent.
- * 3. The skillsMetadata and memoryContents keys are automatically excluded from subagent output
- *    to prevent parent state from leaking to child agents. Each agent loads its own skills/memory
- *    independently based on its middleware configuration.
+ * State keys excluded when passing state to subagents and when returning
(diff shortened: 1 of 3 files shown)


libs/deepagents/src/middleware/subagents.ts
```diff
@@ -560,6 +555,7 @@ function createTaskTool(options: {
       const currentState = getCurrentTaskInput<Record<string, unknown>>();
       const subagentState = filterStateForSubagent(currentState);
       subagentState.messages = [new HumanMessage({ content: description })];
+      subagentState._summarizationSessionId = `session_${crypto.randomUUID().substring(0, 8)}`;
```
hntrl, line 558: what do we need this for?

thushanth-bengre-langchain: To fix the issue open-swe found.

hntrl (approved):

PR #752, opened 2026-08-15
feat(deepagents): introduce ForkedSubAgent type for subagent conversation forking

thushanth-bengre-langchain:
## Summary

Replaces the `mode` field on `SubAgent` with a new `ForkedSubAgent` type for opt-in conversation forking. A `ForkedSubAgent` has no `systemPrompt` of its own — it always inherits the parent's message history and exact system prompt, plus mirrored middleware when its model matches the parent's. `CompiledSubAgent` keeps its existing `mode: "handoff" | "fork"` field, unchanged.

## Changes

- New `ForkedSubAgent` type — same shape as `SubAgent` minus `systemPrompt`; always forks.
- `SubAgent` no longer has a `mode` field — always isolated (same as before this feature existed).
- `CompiledSubAgent.mode` unchanged.
- Mirrored middleware (memory/custom) only added when the model matches the parent's; the system prompt itself is inherited unconditionally either way.
- Updated fork test coverage in `subagent.test.ts` for the new type.

### DIFF
--- a/libs/deepagents/src/middleware/subagents.ts
+++ b/libs/deepagents/src/middleware/subagents.ts
@@ -283,8 +283,8 @@ export interface ForkedSubAgent extends SubAgentBase {
 
 export function isForkedSubAgent(value: unknown): value is ForkedSubAgent {
   if (typeof value !== "object" || value == null) return false;
-  if ("mode" in value && value.mode !== "fork") return false;
-  if ("systemPrompt" in value && value.systemPrompt !== undefined) return false;
+  if (!("mode" in value)) return false;
+  if (value.mode !== "fork") return false;
   return true;
 }
 
--- a/libs/deepagents/src/middleware/subagents.ts
+++ b/libs/deepagents/src/middleware/subagents.ts
@@ -43,8 +43,8 @@ export const DEFAULT_SUBAGENT_PROMPT =
 
 /**
  * State keys excluded when passing state to subagents and when returning
- * updates from subagents. Fork-mode subagents re-inject summarization keys
- * after reconstructing the effective message view.
+ * updates from subagents. Summarization keys are excluded because their
+ * cutoffIndex is only valid against the message list it was computed from.
  */
 const EXCLUDED_STATE_KEYS = [
   "messages",
@@ -107,16 +107,11 @@ export interface CompiledSubAgent<
 }
 
 /**
- * Specification for a subagent that can be dynamically created.
- *
- * When using `createDeepAgent`, subagents automatically receive a default middleware
- * stack (filesystemMiddleware, summarizationMiddleware, etc.) before any custom
- * `middleware` specified in this spec. Add `todoListMiddleware` explicitly to opt in.
+ * Fields shared by both {@link SubAgent} and {@link ForkedSubAgent}.
  *
  * Required fields:
  * - `name`: Identifier used to select this subagent in the task tool
  * - `description`: Shown to the model for subagent selection
- * - `systemPrompt`: The system prompt for the subagent
  *
  * Optional fields:
  * - `model`: Override the default model for this subagent
@@ -124,27 +119,13 @@ export interface CompiledSubAgent<
  * - `middleware`: Additional middleware appended after defaults
  * - `interruptOn`: Human-in-the-loop configuration for specific tools
  * - `skills`: Skill source paths for SkillsMiddleware (e.g., `["/skills/user/", "/skills/project/"]`)
- *
- * @example
- * ```typescript
- * const researcher: SubAgent = {
- *   name: "researcher",
- *   description: "Research assistant for complex topics",
- *   systemPrompt: "You are a research assistant.",
- *   tools: [webSearchTool],
- *   skills: ["/skills/research/"],
- * };
- * ```
  */
-export interface SubAgent {
+interface SubAgentBase {
   /** Identifier used to select this subagent in the task tool */
   name: string;
 
   /** Description shown to the model for subagent selection */
   description: string;
-
-  /** The system prompt to use for the agent */
-  systemPrompt: string;
 
   /** The tools to use for the agent (tool instances, not names). Defaults to defaultTools */
   tools?: StructuredTool[];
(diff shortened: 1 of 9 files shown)


libs/deepagents/src/middleware/subagents.ts
```diff
@@ -96,6 +98,14 @@ export interface CompiledSubAgent<
...
   /** The agent instance */
   runnable: TRunnable;
+
+  /**
+   * Context mode. `"fork"` inherits the parent's conversation history
+   * (but not its system prompt — that's baked into the runnable).
+   * `"dynamic"` lets the model choose per call via the task tool's
+   * `mode` argument. `"handoff"` (default) is fully isolated.
+   */
+  mode?: "handoff" | "fork" | "dynamic";
```
hntrl: for POC can we just do "handoff" | "fork"

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -220,6 +230,32 @@ export interface SubAgent {
...
+  specModel: LanguageModelLike | string | undefined,
+  parentModel: LanguageModelLike | string,
+  parentSystemPrompt: string | SystemMessage | null | undefined,
+): boolean {
+  const isForkable = mode === "fork" || mode === "dynamic";
+  const modelMatchesParent =
+    specModel === undefined || specModel === parentModel;
+  const hasParentSystemPrompt =
+    parentSystemPrompt != null && parentSystemPrompt !== "";
+  return isForkable && modelMatchesParent && hasParentSystemPrompt;
```
hntrl: we shouldn't be gating on supervisor + child models being the same

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -465,12 +523,34 @@ function getSubagents(options: {
...
+        subagentSystemPrompts[resolvedSpec.name] = resolvedSpec.systemPrompt;
+      }
+      const systemPromptOverride =
+        resolvedSpec.mode === "fork" && cacheAligned
+          ? (parentSystemPrompt ?? undefined)
+          : undefined;
+
+      agents[agentParams.name] = createSubAgent(resolvedSpec, {
+        systemPromptOverride,
+      });
```
hntrl:
```suggestion
      resolvedSpec.systemPrompt =
        resolvedSpec.mode === "fork" && cacheAligned
          ? (parentSystemPrompt ?? undefined)
          : undefined;

      agents[agentParams.name] = createSubAgent(resolvedSpec);
```

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -536,7 +646,11 @@ function createTaskTool(options: {
 
   return tool(
     async (
-      input: { description: string; subagent_type: string },
+      input: {
+        description: string;
+        subagent_type: string;
+        mode?: "handoff" | "fork";
+      },
```
hntrl:
```suggestion
      input,
```

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -465,12 +531,32 @@ function getSubagents(options: {
         ],
         interruptOn: agentParams.interruptOn ?? defaultInterruptOn ?? undefined,
       };
-      agents[agentParams.name] = createSubAgent(resolvedSpec);
+
+      const systemPromptOverride = isCacheAlignedForkSpec({
```
hntrl: do we need a special option for systemPromptOverride? any reason to not just set `spec.systemPrompt`?

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -550,12 +645,37 @@ function createTaskTool(options: {
...
+          : trimmed;
+        const ownSystemPrompt = subagentSystemPrompts[subagent_type];
+        const content = ownSystemPrompt
+          ? `${ownSystemPrompt}\n\n${description}`
+          : description;
+        subagentState.messages = [...effective, new HumanMessage({ content })];
+        // Not re-adding _summarizationEvent: cutoffIndex already applied above.
+        subagentState._summarizationSessionId =
+          (currentState._summarizationSessionId as string | undefined) ??
+          `session_${crypto.randomUUID().substring(0, 8)}`;
```
hntrl: dont we want subagents to also carry their own session id

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -550,12 +645,37 @@ function createTaskTool(options: {
...
+        const event = currentState._summarizationEvent as
+          | SummarizationEvent
+          | undefined;
+        const effective = event
+          ? [event.summaryMessage, ...trimmed.slice(event.cutoffIndex)]
+          : trimmed;
+        const ownSystemPrompt = subagentSystemPrompts[subagent_type];
+        const content = ownSystemPrompt
+          ? `${ownSystemPrompt}\n\n${description}`
+          : description;
```
hntrl: weren't we just going to disallow the `systemPrompt` param when creating a forked subagent? models weight system messages differently then user messages so my hunch is in any case we shouldn't implicitly put the contents in a human message

libs/deepagents/src/middleware/subagents.ts
thushanth-bengre-langchain: Subagents `systemPrompt` is a string but `parentSystemPrompt` can be a `SystemMessage`.
Additionally, for non-cache aligned flow, systemPromptOverride can be undefined. If we directly set it to to systemPrompt, that can have unwanted behavior in multiple places.

libs/deepagents/src/middleware/subagents.ts
thushanth-bengre-langchain: This is a bug. We don't want to use parents session id's.
Fixed this.

libs/deepagents/src/middleware/subagents.ts
hntrl:
```suggestion
  systemPrompt: string | SystemMessage;
```

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -259,10 +298,11 @@ export interface SubAgent {
...
  */
-export const GENERAL_PURPOSE_SUBAGENT: Pick<
-  SubAgent,
-  "name" | "description" | "systemPrompt"
-> = {
+export const GENERAL_PURPOSE_SUBAGENT: {
+  name: string;
+  description: string;
+  systemPrompt: string;
+} = {
```
hntrl:
```suggestion
export const GENERAL_PURPOSE_SUBAGENT: Pick<
  SubAgent,
  "name" | "description" | "systemPrompt"
> = {
```

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -550,11 +645,30 @@ function createTaskTool(options: {
...
+      if (shouldFork) {
+        const trimmed = stripInFlightAIMessage(
+          (currentState.messages as BaseMessage[]) ?? [],
+        );
+        const event = currentState._summarizationEvent as
+          | SummarizationEvent
+          | undefined;
+        const effective = event
+          ? [event.summaryMessage, ...trimmed.slice(event.cutoffIndex)]
+          : trimmed;
```
hntrl: I think it would be better if we moved `getEffectiveMessages` from middleware/summarization.ts to the top level, exported it, then used it here

DRY and all that

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -447,16 +502,41 @@ function getSubagents(options: {
...
+        interruptOn: agentParams.interruptOn ?? defaultInterruptOn ?? undefined,
+      };
+      agents[agentParams.name] = createSubAgent(resolvedSpec);
+      specsByName[agentParams.name] = resolvedSpec;
     } else {
+      // ForkedSubAgent — always forks, always inherits the parent's exact
+      // system prompt directly; there's no own prompt to fall back to.
       const resolvedSpec: SubAgent = {
         ...agentParams,
+        systemPrompt: parentSystemPrompt ?? "",
```
hntrl: why do we need to thread parentSystemPrompt through here? When it comes time to call `task.invoke` those system message objects are already retained in state

hntrl: if we include this I think that means when we read the current message array from state the system prompts will be duplicated

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -222,6 +211,56 @@ export interface SubAgent {
...
+ * @example
+ * ```typescript
+ * const researcher: ForkedSubAgent = {
+ *   name: "researcher",
+ *   description: "Continues the current investigation with full context",
+ *   tools: [webSearchTool],
+ * };
+ * ```
+ */
+export interface ForkedSubAgent extends SubAgentBase {}
```
hntrl: I think we should keep mode

```suggestion
export interface ForkedSubAgent extends SubAgentBase {
  mode?: "fork";
}
```

libs/deepagents/src/middleware/subagents.ts
thushanth-bengre-langchain: Checked this. `state` does not contain parent system prompt. It only has AI/Human/Tool messages. 
That's why we need to thread this explicitly.

hntrl: hurmm i'm surprised

hntrl (dismissed):

hntrl (dismissed):

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -222,6 +214,80 @@ export interface SubAgent {
...
+   * the parent's conversation history and system prompt.
+   */
+  mode?: "fork";
+}
+
+export function isForkedSubAgent(value: unknown): value is ForkedSubAgent {
+  if (typeof value !== "object" || value == null) return false;
+  if ("mode" in value && value.mode !== "fork") return false;
+  if ("systemPrompt" in value && value.systemPrompt !== undefined) return false;
+  return true;
```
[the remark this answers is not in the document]
hntrl: fixed

PR #766, opened 2026-08-19
fix(deepagents): return raw file content from BaseSandbox.read()

christian-bromann:
## Summary

The sandbox awk command was numbering lines, then read_file numbered them again. Other backends already return raw content.

- `BaseSandbox.read()` was emitting `cat -n` style line numbers from awk, and `read_file` numbered the same content again, so the model saw a doubled gutter.
- The sandbox now returns raw file content like the other backends; the tool still owns the single gutter. Python already works this way (`BaseSandbox` paginates raw text, middleware calls `format_content_with_line_numbers` once).

## Test plan
- [x] `pnpm vitest run` in `libs/deepagents` (1384 passed)
- [x] Regression tests for raw sandbox output + single `read_file` gutter
- [x] Changeset added for affected package(s)

### DIFF
--- a/libs/deepagents/src/backends/sandbox.test.ts
+++ b/libs/deepagents/src/backends/sandbox.test.ts
@@ -64,15 +64,9 @@ class MockSandbox extends BaseSandbox {
             truncated: false,
           };
         }
-        // Mirror awk's printf format so a gutter added back to the command
-        // shows up here rather than being silently normalized away.
-        const numbered = command.includes("%6d");
-        const output = content
-          .split("\n")
-          .map((line, i) =>
-            numbered ? `${String(i + 1).padStart(6)}\t${line}` : line,
-          )
-          .join("\n");
+        // Always return the page as raw lines. Numbering belongs to the
+        // read_file tool; this mock mirrors the BaseSandbox contract.
+        const output = content.split("\n").join("\n");
         return { output, exitCode: 0, truncated: false };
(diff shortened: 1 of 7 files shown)


libs/deepagents/src/backends/sandbox.test.ts
```diff
@@ -64,9 +64,14 @@ class MockSandbox extends BaseSandbox {
...
-        const output = lines
-          .map((line, i) => `     ${i + 1}\t${line}`)
+        // Mirror awk's printf format so a gutter added back to the command
+        // shows up here rather than being silently normalized away.
+        const numbered = command.includes("%6d");
+        const output = content
+          .split("\n")
+          .map((line, i) =>
+            numbered ? `${String(i + 1).padStart(6)}\t${line}` : line,
+          )
```
hntrl: this feels very system dependent? e.g. the gutter being 6 spaces big. i've got no evidence to the contrary but we might just be able to look for the common prefix

christian-bromann: Agreed — dropped the `%6d` / `padStart(6)` coupling.

The mock now always returns raw lines (matching the backend contract). The regression just asserts the awk `printf` emits the line body (`"%s\n"`) and does not include any `%Nd` number field, so it does not care about gutter width.

hntrl (approved):

PR #800, opened 2026-08-28
fix(deepagents): backport Python's fork cache-parity and recursion-safety fixes

thushanth-bengre-langchain:
## Summary

This PR backports a set of fixes that landed later in Python
`deepagents`'s equivalent feature (langchain-ai/deepagents#5714)

## What's included

- **Dynamic system-message capture and replay** — a fork now gets the parent's actual composed system message (skills, memory, etc.), not a stale static copy, so the prompt cache can hit.
- **Identity-preserving preamble** — tells a fork it's continuing a prior conversation, not receiving a fresh request.
- **Recursion refusal** — a fork's task tool refuses re-delegation instead of recursing.
- **Tool-description hint for the parent** — marks forked subagents in the task tool listing so the parent knows they don't need context restated.
- **Memory-middleware parity fix** — forks no longer get memoryMiddleware mirrored in, matching Python.

## Testing

- New unit tests: skills-injected system message replay into a fork,
  rejection of `skills` on a `ForkedSubAgent` at construction, a fork's own
  attempt to delegate again being refused rather than recursing, and
  `parentSystemMessageMiddleware` only being installed when a declarative
  fork is actually present.

### DIFF
--- a/.changeset/forked-subagent-conversation.md
+++ b/.changeset/forked-subagent-conversation.md
@@ -1,0 +1,5 @@
+---
+"deepagents": minor
+---
+
+Add experimental conversation forking for subagents (`mode: "fork"`), inheriting the parent's message history and system prompt for prompt-cache reuse.
--- a/.changeset/forked-subagent-conversation.md
+++ b/.changeset/forked-subagent-conversation.md
@@ -1,5 +1,5 @@
 ---
-"deepagents": minor
+"deepagents": patch
 ---
 
 Add experimental conversation forking for subagents (`mode: "fork"`), inheriting the parent's message history and system prompt for prompt-cache reuse.
--- a/libs/deepagents/src/middleware/subagents.ts
+++ b/libs/deepagents/src/middleware/subagents.ts
@@ -465,7 +465,7 @@ function stripInFlightAIMessage(messages: BaseMessage[]): BaseMessage[] {
 }
 
 const ParentSystemMessageStateSchema = z.object({
-  _deepagentsParentSystemMessage: z.instanceof(SystemMessage).optional(),
+  [PARENT_SYSTEM_MESSAGE_KEY]: z.instanceof(SystemMessage).optional(),
 });
 
 // Captures the parent's fully-resolved system message into state so a fork can replay it verbatim.
@@ -476,7 +476,7 @@ export function createParentSystemMessageMiddleware(): AgentMiddleware {
     wrapModelCall: async (request, handler) => {
       await handler(request);
       return new Command({
-        update: { _deepagentsParentSystemMessage: request.systemMessage },
+        update: { [PARENT_SYSTEM_MESSAGE_KEY]: request.systemMessage },
       });
     },
   });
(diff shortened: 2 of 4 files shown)


.changeset/forked-subagent-conversation.md
```diff
@@ -0,0 +1,5 @@
+---
+"deepagents": minor
```
hntrl:
```suggestion
"deepagents": patch
```

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -414,6 +453,55 @@ function stripInFlightAIMessage(messages: BaseMessage[]): BaseMessage[] {
   return hasPendingToolCalls ? messages.slice(0, -1) : messages;
 }
 
+const ParentSystemMessageStateSchema = z.object({
+  _deepagentsParentSystemMessage: z.instanceof(SystemMessage).optional(),
```
hntrl: think you can tie to the constant
```suggestion
  [PARENT_SYSTEM_MESSAGE_KEY]: z.instanceof(SystemMessage).optional(),
```

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -54,6 +63,8 @@ const EXCLUDED_STATE_KEYS = [
   "memoryContents",
   "_summarizationEvent",
   "_summarizationSessionId",
+  PARENT_SYSTEM_MESSAGE_KEY,
+  FORKED_CONTEXT_KEY,
```
hntrl, line 67: dont we want this to propagate to the subagent? or is this just on the return state

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -414,6 +453,55 @@ function stripInFlightAIMessage(messages: BaseMessage[]): BaseMessage[] {
...
+      if (parentMessage != null) {
+        return handler({ ...request, systemMessage: parentMessage });
+      }
+      return handler(request);
+    },
+  });
+}
+
+const ForkedContextStateSchema = z.object({
+  _deepagentsForkedContext: z.boolean().optional(),
```
hntrl: use constant here too

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -493,6 +583,7 @@ function getSubagents(options: {
     subagents,
     generalPurposeAgent,
     parentSystemPrompt = null,
+    mirroredTaskTool,
```
hntrl, line 597: why do we have to propagate `mirroredTaskTool` through the chain?

hntrl: disregard, I think this is kind of awkward but I know why we need it

libs/deepagents/src/middleware/subagents.ts
thushanth-bengre-langchain: Done.

libs/deepagents/src/middleware/subagents.ts
thushanth-bengre-langchain: This is for the return state.

hntrl (approved):

Issue #646, opened 2026-07-05
InvalidUpdateError: LastValue can only receive one value per step — parallel task calls when both parent and SubAgent spec use modelCallLimitMiddleware


nobodyme:
## Description

When the **parent agent** and a **`SubAgent` spec** both use `modelCallLimitMiddleware` (a natural config: cap the main agent's run AND cap each subagent), two parallel `task` calls crash the whole parent run with `INVALID_CONCURRENT_GRAPH_UPDATE` — and parallel `task` calls are exactly what the built-in task-tool description encourages ("Launch multiple agents concurrently whenever possible").

Looks like an unreported variant of the LastValue-collision family from #65 / #144 (and the afterModel variant in #480): those were fixed for deepagents' own channels (e.g. `files` merges now), but state written by *user-supplied middleware in a subagent spec* still round-trips into the parent graph unfiltered.

Reproduces on latest (`deepagents@1.10.5`, `langchain@1.5.2`, `@langchain/langgraph@1.4.7`) and on `1.10.2`/`1.4.4`/`1.3.3`.  Minimal deterministic repro below (no API keys — a scripted model always emits two parallel task calls).

## Reproduction

```js
// repro.mjs — node repro.mjs
import { createDeepAgent } from "deepagents";
import { modelCallLimitMiddleware } from "langchain";
import { BaseChatModel } from "@langchain/core/language_models/chat_models";
import { AIMessage } from "@langchain/core/messages";

class ScriptedModel extends BaseChatModel {
  calls = 0;
  constructor() { super({}); }
  _llmType() { return "scripted"; }
  bindTools() { return this; }
  async _generate() {
    this.calls++;
    const msg = this.calls === 1
      ? new AIMessage({
          content: "Exploring two areas in parallel.",
          tool_calls: [
            { id: "t1", name: "task", args: { description: "Find where user auth lives", subagent_type: "explorer" }, type: "tool_call" },
            { id: "t2", name: "task", args: { description: "Find where notification prefs are stored", subagent_type: "explorer" }, type: "tool_call" },
          ],
        })
      : new AIMessage({ content: "FINDINGS: none" });
    return { generations: [{ message: msg, text: "" }] };
  }
}

const agent = createDeepAgent({
  model: new ScriptedModel(),
  systemPrompt: "You are a test agent.",
  // Parent caps its own run — the documented pattern. This declares the
  // threadModelCallCount channel on the parent graph.
[...]

PR #842, opened 2026-09-08
fix(deepagents): scope subagent call-count state to its own agent

byhow:
## Summary

`modelCallLimitMiddleware` and `toolCallLimitMiddleware` declare their counters as plain values, so LangGraph gives them single-writer `LastValue` channels. The `task` tool returned a subagent's whole final state to the parent, so those counters crossed the graph boundary in both directions:

- Parallel delegation crashed - two subagents finishing in one super-step both wrote the channel, failing the run with `InvalidUpdateError` in #646 .
- Serial delegation silently bypassed the limit - the subagent's `afterAgent` reset landed in the parent and rewound its budget, so a parent with `runLimit: 3` made 10 model calls, and its `threadModelCallCount` read 6 after 2 real calls.

Forks are affected with no subagent-side configuration, since a fork inherits the parent's middleware.

This excludes the four counter keys from both subagent state filters — `filterStateForSubagent`, which covers isolated-subagent input and every subagent's returned update, and `filterStateForFork`, which covers fork input — so each agent counts only its own calls. Deliberately shared state such as `files` still propagates.

## Known limitation

Matching by name misses other middleware with plain state, reachable via a `mode: "fork"` subagent plus parent-side custom middleware — including the opt-in `createAgentMemoryMiddleware` and `createCompletionCallbackMiddleware`. Two tests assert that current behaviour so it stays visible. The general fix, deriving keys from each subagent's middleware `stateSchema` and forwarding only `ReducedValue` channels, is deferred to a follow-up.


libs/deepagents/src/middleware/subagents.state-isolation.test.ts
hntrl, line 1: nit pick: can we coalesce these into existing test files? also can we add integration tests?

libs/deepagents/src/middleware/subagents.state-isolation.test.ts
```diff
@@ -0,0 +1,439 @@
...
+    return {
+      generations: [
+        {
+          message,
+          text: typeof message.content === "string" ? message.content : "",
+        },
+      ],
+    };
+  }
+}
```
hntrl: langchain already exports a few utilities for chat model mocking

libs/deepagents/src/middleware/subagents.ts
```diff
@@ -49,6 +49,27 @@ const FORKED_CONTEXT_KEY = "_deepagentsForkedContext";
 const FORK_RECURSION_REFUSAL =
   "You are a subagent and cannot delegate to another subagent. Complete this task yourself instead of calling this tool again.";
 
+/**
+ * Per-graph call accounting owned by the upstream limit middlewares. Upstream
+ * declares these as plain values, so LangGraph gives them single-writer
+ * `LastValue` channels: forwarding them crashed parallel delegation and let a
+ * subagent's reset overwrite the parent's budget (#646). Each agent counts its
```
hntrl: nit: I wouldn't include github refs in public docs
```suggestion
 * subagent's reset overwrite the parent's budget. Each agent counts its
```

libs/deepagents/src/middleware/subagents.state-isolation.test.ts
```diff
@@ -0,0 +1,439 @@
...
+    model: new ScriptedChatModel(() => finalMessage("subagent done")),
+    ...overrides,
+  };
+}
+
+const invoke = (agent: any, recursionLimit = 40) =>
+  agent.invoke(
+    { messages: [new HumanMessage("go")] },
+    { recursionLimit },
+  ) as Promise<Record<string, unknown>>;
```
hntrl: I think its fine to repeat ourselves in these test files -- adding these utilities is another layer of indirection thats hard to reason about

libs/deepagents/src/middleware/subagents.state-isolation.test.ts
byhow: Yes can do

libs/deepagents/src/middleware/subagents.state-isolation.test.ts
byhow: Switched to `SequentialFakeChatModel`, the existing helper in `subagent.test.ts` that wraps `FakeListChatModel` from `@langchain/core/utils/testing`. Moving the tests into that file meant I could reuse it rather than carry my own.

For the record on why the bespoke one existed: plain `FakeListChatModel` restarts its response list on `bindTools()`, which breaks multi-turn flows. `SequentialFakeChatModel` already solves that with a shared counter, and its `counter` option also gave me a clean way to assert the parent's model-call count in the run-limit test.

libs/deepagents/src/middleware/subagents.ts
byhow: Sounds good, applied the changes

libs/deepagents/src/middleware/subagents.state-isolation.test.ts
byhow: Sounds good, each test now spells out its own agent, subagent spec and assertions inline

hntrl (approved):

hntrl:
for posterity -- this is by no means the end-state fix. I think eventually we'll want to put StateSchema api under a microscope and figure out how we enable things like input/output filtering on state keys specifically

PR #145, opened 2026-01-26
fix(deepagents): use new StateSchema class

christian-bromann:
This PR fixes a LangGraph `LastValue` channel error that occurs when parallel subagents update state simultaneously. The fix migrates state schemas to use LangGraph's new `StateSchema` class with `ReducedValue` to enable proper concurrent state updates.

**Problem:** When skills or filesystem middleware is enabled and multiple subagents run in parallel, each subagent returns state updates (e.g., `skillsMetadata`, `files`). Since these fields used implicit `LastValue` semantics (no custom reducer), LangGraph fails with:

```
Invalid update for channel "skillsMetadata" with values [[...],[...],[...]]:
LastValue can only receive one value per step.
```

**Solution:** Use `StateSchema` with `ReducedValue` and custom reducers that properly merge concurrent updates:

- `skillsMetadataReducer`: Merges skill arrays, deduplicating by name (later values override earlier)
- `fileDataReducer`: Merges file records with support for deletions (null values remove files)

closes #142
fixes #143
fixes #144
fixes #122

PR #784, opened 2026-08-22
feat: add optional You.com search integration

mouse-value-add:
## Overview

This PR adds an optional You.com web search integration to Deep Agents, providing an alternative search provider alongside the existing Tavily integration in the research examples.

## What's Changed

- **New You.com search tool** (`tools/youcom-search.ts`) with full TypeScript support
- **Enhanced research agent** that includes both Tavily and You.com search options
- **Standalone example** (`youcom-search-agent.ts`) demonstrating You.com-only usage
- **Comprehensive documentation** (`README-youcom.md`) with setup, usage, and troubleshooting

## Key Features

✅ **Flexible authentication**: Works with or without API key (keyless mode supported)
✅ **Error handling**: Graceful handling of rate limits, auth errors, and network issues  
✅ **TypeScript integration**: Full type safety with Zod schema validation
✅ **LangChain compatibility**: Follows standard LangChain tool patterns
✅ **Zero breaking changes**: Pure additive integration, existing code unchanged

### Basic Setup (Keyless)
```typescript
import { youcomSearch } from "./tools/youcom-search.js";

const agent = createDeepAgent({
  model: new ChatAnthropic({ model: "claude-sonnet-4-20250514" }),
  tools: [youcomSearch], // Works without API key
});
```

### Enhanced Setup (With API Key)
```bash
# Optional: Add to .env for enhanced features
YDC_API_KEY=your_api_key
```

## Integration Approach

This integration follows Deep Agents' existing patterns:
- Uses the same tool structure as the Tavily integration
- Maintains consistency with existing search tool interfaces
- Provides multiple search provider options for research workflows
- Includes proper error handling and user feedback

## Testing

The integration has been tested for:
- ✅ TypeScript compilation and type safety
- ✅ Tool parameter validation with Zod schemas
- ✅ Error handling for various API scenarios
- ✅ Integration with Deep Agents' tool system
- ✅ Keyless and authenticated operation modes

## Why This Integration

You.com's search API provides:
- High-quality, curated search results
- Strong source attribution and citations  
- Flexible authentication options
- Complementary capabilities to existing search providers

This gives Deep Agents users more choice in search providers and can improve research quality by combining multiple search sources.

## Related Issue

Tracking: https://github.com/youdotcom-oss/integration-tracking/issues/212

PR #860, opened 2026-09-14
feat(deepagents): reload skills when `skillsMetadata` is set to `null`

### LINKED ISSUE #562: skills gets cached in checkpoint and does not reflect after new skill is uploaded
Im facing issue when i deploy new skill inside a local folder but when i ask deepagents what skills you have access to it gives a previous copy from past thread id which probably is checkpointed in postgres db so what is the recommened way to avoid this problem

richardscarrott:
## Summary

With a checkpointer, a thread loads skills once and keeps them in `skillsMetadata`. Skills added, edited or deleted in the backend after that never reach the model. There was no supported reset. `updateState` with `null` failed validation, the reducer ignored `[]`, and a new list merged by name, so deleted skills stayed.

Set `skillsMetadata` to `null` between runs, and the next run reloads every source and replaces the stored list:

```ts
await agent.updateState(config, { skillsMetadata: null });
// or as part of the next run's input
await agent.invoke({ messages, skillsMetadata: null }, config);
```

- `skillsMetadata` is now a plain `.nullish()` field with no reducer and no default.
- `beforeAgent` loads when the value is `null` or missing. A stored `[]` now means the sources had no skills, so those threads stop listing the backend on every run. That includes existing checkpoints that store `[]`.
- `skillsMetadataReducer` is gone. No package entry point exported it.

This is the same contract as langchain-ai/deepagents#6288, with `null` in place of `None`.

Closes #562

## Why removing the reducer is safe

- #145 added the reducer and the `[]` default because parallel subagents returned `skillsMetadata` and hit "LastValue can only receive one value per step". The reducer ignored empty updates because those subagents returned empty lists.
- #160 switched the load check to "non-empty list", because the `[]` default made a `!= null` check always true.
- #187 excluded `skillsMetadata` from subagent state in both directions. That removed the reason for the reducer, but the reducer stayed.
- Nothing else writes the key. Forks still receive the parent's list, and no subagent hands a list back.
- A new test sends a general-purpose `task` call and a fork in one message and checks the parent's list afterwards. With `skillsMetadata` taken out of `EXCLUDED_STATE_KEYS`, it fails with `INVALID_CONCURRENT_GRAPH_UPDATE`. That test keeps the removal safe if someone edits the exclusion list.

## Rolling back after a reset

If a thread is reset and deepagents is then downgraded before that thread runs again, `agent.invoke`/`agent.stream` on the older version throws `Validation failed for field "skillsMetadata"`, because older versions don't accept `null`. To recover, run `agent.updateState(config, { skillsMetadata: [] })` on the older version. LangGraph API server deployments aren't affected.

Docs: langchain-ai/docs#6043


.changeset/skills-reload-on-null.md
```diff
@@ -0,0 +1,17 @@
+---
+"deepagents": minor
+---
+
+feat(deepagents): reload skills when `skillsMetadata` is set to `null`
```
hntrl, line 5: do we want this to be patch?

libs/deepagents/src/middleware/skills.ts
```diff
@@ -218,49 +218,11 @@ export const SkillMetadataEntrySchema = z.object({
...
-  // Merge by skill name (later values override earlier ones)
-  const merged = new Map<string, SkillMetadataEntry>();
-  for (const skill of current) {
-    merged.set(skill.name, skill);
-  }
-  for (const skill of update) {
-    merged.set(skill.name, skill);
-  }
-  return Array.from(merged.values());
-}
```
hntrl, line 250: why are we dropping the reducer? are we losing any behaviors because of this? naively I think using `Map` helps us deterministically order skill entries?

.changeset/skills-reload-on-null.md
richardscarrott: I felt the removal of the `skills_metadata` reducer was enough of a breaking change that it probably should be a minor?

libs/deepagents/src/middleware/skills.ts
richardscarrott: So the reducer was introduced here https://github.com/langchain-ai/deepagentsjs/pull/145 because:

> When skills or filesystem middleware is enabled and multiple subagents run in parallel, each subagent returns state updates (e.g., skillsMetadata, files). Since these fields used implicit LastValue semantics (no custom reducer), LangGraph fails [with "Invalid update for channel"]

However a subsequent PR https://github.com/langchain-ai/deepagentsjs/pull/187 added `skillsMetadata` to `EXCLUDED_STATE_KEYS` which meant the scenario from #145 is no longer relevant (as it's filtered out from the state update)

Ultimately there's no code path in deepagents which calls the reducer w/ a prev and next value that differ.

However, it does impact user `skillsMetadata` writes as without the reducer it'll replace the old skills rather than merge (or if parallel tools both try to write it'll error with "Invalid update for channel") which is why I thought a `minor` version bump was better than `patch`.

Also, for parity the python SDK doesn't have this reducer and instead it uses the default `LastValue`.

---------------

re: ordering, it looks like that's determined by a) the backend and b) the source stacking / deduping which still behaves as before.

.changeset/skills-reload-on-null.md
```diff
@@ -0,0 +1,20 @@
...
+
+Skills are loaded once per thread and kept in state, so a long-lived thread never saw skills added, edited or deleted after its first run. Setting `skillsMetadata` to `null` now makes the next model call re-read every source and replace the stored list. A thread whose sources hold no skills also counts as loaded, so it stops listing the backend on every run — including threads checkpointed by earlier versions, which store `[]` and need an explicit `null` to reload.
+
+Loading moved from `beforeAgent` to `beforeModel`, matching Python's `SkillsMiddleware`, so a reload is served by the next model call rather than the next run. Two consequences:
```
hntrl:
```suggestion
Loading moved from `beforeAgent` to `beforeModel`, so a reload is served by the next model call rather than the next run. Two consequences:
```

.changeset/skills-reload-on-null.md
```diff
@@ -0,0 +1,20 @@
...
+
+- A middleware of your own can invalidate mid-run, from `afterModel`, and the following model call sees the fresh list.
+- With a `StateBackend`, a load now sees `files` written during the run, so a `SKILL.md` the agent just wrote is picked up.
+
+One gap remains: a `jumpTo: "model"` from an `afterModel` hook routes straight to the model node and bypasses `beforeModel`, so a reload pending at that moment is not served until the iteration after. `humanInTheLoopMiddleware` jumps this way when a tool call is rejected, so a reload requested during a review interrupt is skipped for that one model call.
```
hntrl: maybe not appropriate for a changeset?
```suggestion
```

libs/deepagents/src/middleware/skills.ts
```diff
@@ -818,15 +823,13 @@ export function createSkillsMiddleware(options: SkillsMiddlewareOptions) {
...
-      // Already loaded for this thread (same run, a later turn, or a
-      // restored checkpoint) - keep what state holds.
-      if (stateHasSkills) {
+    async beforeModel(state) {
+      // What state holds for this thread:
+      // - missing, `undefined` or `null`: not loaded, or the caller reset it.
+      //   Load every source.
+      // - `[]`: loaded, and the sources contain no skills. Keep it.
+      // - a non-empty list: loaded. Keep it.
+      if (state.skillsMetadata != null) {
```
hntrl: man I hate javascript syntax sometimes, can we make this more explicit?


```suggestion
      if (state.skillsMetadata !== null && state.skillsMetadata !== undefined) {
```

libs/deepagents/src/agent.ts
```diff
@@ -551,6 +555,9 @@ export function createDeepAgent<
    * This ensures InferMiddlewareStates captures state from both sources
    */
   type AllMiddleware = readonly [
+    // Middleware mounted only when its option is present (see the runtime
+    // `skillsMiddleware` spread above) — keeps state in sync with the options.
+    ...InferOptionalMiddleware<TSkills>,
```
hntrl, line 560: our types are scary sometimes

hntrl: (no action needed)

.changeset/skills-reload-on-null.md
richardscarrott: Ah soz yeah, agree. Claude getting too chatty.

hntrl (approved):

richardscarrott:
Closing in favour of #862, which is this branch pushed to a branch on this repo at the same commit (`bd23287`) — identical diff, nothing rebased.

CodeQL here is on GitHub's default setup, which doesn't analyse fork PR refs, so this PR had 0 CodeQL runs while the `main` ruleset requires CodeQL results. That left it stuck on `Code scanning is waiting for results from CodeQL` permanently rather than pending. #862 picked up a CodeQL run immediately.

Review discussion here stays the reference; thanks @hntrl for the review. Worth noting this blocks every fork PR, not just this one — #809, #807 and #784 are all in the same state.

PR #861, opened 2026-09-14
feat(deepagents,quickjs): update `read_file` output formatting

thushanth-bengre-langchain:
## Summary
- `read_file` now emits a single `@@ lines A-B[ of T] | ... @@` status header above unmodified source, instead of a per-line gutter. Folds pagination/truncation into terse header fields (adds mid-line truncation and negative-offset clamp disclosures, matching Python).
- Fixes quickjs's `stripLineNumbers`, which relied on the old gutter and would've silently broken; also tightens its header regex to avoid false-positives on real `@@ ... @@` content.
- Ported from [langchain-ai/deepagents#5648](https://github.com/langchain-ai/deepagents/pull/5648).


.changeset/plain-file-headers.md
hntrl, line 1: can we give some snippets like in the python PR description? just demonstrating before and after of the formatting

libs/providers/quickjs/src/session.ts
```diff
@@ -49,6 +49,15 @@ export const DEFAULT_MAX_SUBAGENT_CONCURRENCY = 32;
 
 const LINE_NUMBER_RE = /^\s*\d+(?:\.\d+)?\t/;
 
+const READ_STATUS_HEADER_RE =
+  /^@@ lines (\d+)-(\d+)(?: of \d+)?(?: \| .*)? @@$/;
+
+// A bracketed notice line, e.g. `[Output was truncated due to size limits...]`.
+const READ_NOTICE_RE = /^\[.*\]$/;
```
hntrl, line 56: I don't love that we have to rely on regex matching to surface tool information, is there any smarter way we can carry this on the ToolMessage instead of regex matching? (would timebox this, its how we do it in py but I'm wondering if theres something smarter we can do)

libs/providers/quickjs/src/session.ts
```diff
@@ -129,26 +135,34 @@ function extractToolText(result: unknown): string {
...
   const lines = text.split("\n");
   if (lines.length === 0) {
     return text;
   }
 
+  const headerIndex = lines
+    .slice(0, MAX_READ_NOTICE_LINES + 1)
+    .findIndex((l) => READ_STATUS_HEADER_RE.test(l));
+  if (headerIndex !== -1) {
+    return lines.slice(headerIndex + 1).join("\n");
```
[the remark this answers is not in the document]
thushanth-bengre-langchain: Added a regression test in commit `d2fdc26`.
Added `READ_NOTICE_RE` check requiring every line before a candidate header to be a bracketed notice `[...]`

libs/providers/quickjs/src/session.ts
thushanth-bengre-langchain: One other way to do this is we can add this as a metadata. Let me explore that.

.changeset/plain-file-headers.md
thushanth-bengre-langchain: Updated Changeset

libs/providers/quickjs/src/session.ts
thushanth-bengre-langchain: Carrying artifact metadata on every tool call is one way we can handle this. But that changes the shape of every toolcall structure, not just `read_file`. 
Much bigger change. imo, not worth doing here.

I tightened the notice-matching regex so a custom tools real content is not mistaken for a header. 

Let me know if you want me to scope out carrying metadata for toolCall.

ccurme: Use of regex here isn't new right? Since we use it to strip the existing numbered gutters.

I do think we need something smarter for code interpreter tools more generally (e.g., allowing tools to specify "machine-readable" content which differs from `.content` and using that in PTC scenarios).

I wouldn't block on this here though since it's not new.

ccurme (approved):
If users upgrade deepagents without upgrading quickjs, we won't strip the new header content; not sure if there's anything we can do about that.

PR #866, opened 2026-09-18
fix(deepagents): close virtual-mode symlink escape in FilesystemBackend

thushanth-bengre-langchain:
Addresses [LC-587](https://linear.app/langchain/issue/LC-587/filesystembackend-virtual-mode-confinement-is-escaped-through-an)

## Summary

- `FilesystemBackend`'s virtual-mode containment check (`resolvePath`) was purely lexical, so a symlink under the sandbox root pointing outside it could escape confinement for `read`/`write`/`edit`/`ls`/`glob`/`grep`/upload/download.
- Adds a real-path containment check, mirroring Python's `_resolve_path` (which already resolves symlinks via `Path.resolve()` before its containment check): resolves the deepest existing path segment for real and rejects only if that escapes the root, so an in-root symlink still works.

Addresses LC-587.

## Test plan

- Added regression tests that fail against the old code (read/write/edit/ls escaping via an intermediate symlink), plus an escaping-symlink-with-not-yet-created-deeper-path edge case and a pin that in-root symlinks still work.
- `pnpm --filter deepagents test` — 1448/1448 passing.


libs/deepagents/src/backends/filesystem.test.ts
```diff
@@ -921,3 +951,137 @@ describe("FilesystemBackend symlink cycle handling", () => {
     },
   );
 });
+
+/**
+ * LC-587: a symlink under the sandbox root pointing outside it passes
```
hntrl:
```suggestion
 * a symlink under the sandbox root pointing outside it passes
```

hntrl (approved):

